PR #1, opened 2015-07-14
Version 1.0 RC

pcarranzav:


Page-:
It would be nice to use bower for the UI JS dependencies rather than committing them to the repo

src/app.coffee
```diff
@@ -0,0 +1,110 @@
...
+					console.log("Captive portal enabled")
+					server = app.listen port, ->
+						console.log("Server listening")
+
+console.log("Starting node connman app")
+manageConnection = (retryCallback) ->
+	connman.init (err) ->
+		if err?
+			console.log(err)
+			return retryCallback(err)
```
Page-: In general I'd rather use promises to avoid the need for this boilerplate on every single callback and just have it once

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM resin/armv7hf-node:0.12.2
+
+RUN apt-get update && apt-get install -y \
+	dropbear \
```
Page-: Can you alphabetise these please

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+	nano \
+	connman \
+	bridge-utils \
+	bind9
+
+COPY ./assets/bind /etc/bind
+
+RUN mkdir -p /app/src
+COPY ./src/package.json /app/src/
+RUN cd /app/src && npm install
```
Page-: `RUN cd /app/src && JOBS=MAX npm install --unsafe-perm --production && npm cache clean`

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+	libdbus-1-dev \
+	libexpat-dev \
+	nano \
+	connman \
+	bridge-utils \
+	bind9
+
+COPY ./assets/bind /etc/bind
+
+RUN mkdir -p /app/src
```
Page-: Can you use `/usr/src/app` instead of `/app` please

src/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
Page-: As a general thing package.json usually goes in the repo root

src/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "name": "resin-wifi-connect",
+  "version": "1.0.0",
+  "repository": "https://github.com/pcarranzav/resin-wifi-connect",
+  "description": "A tool to allow wifi configuration to be set via a captive portal",
+  "scripts": {
+    "start": "coffee app.coffee",
```
Page-: It would help a lot to avoid the need to run the coffee-script compile on the pi, so compile the coffee-script in the Dockerfile and then just call the pre-compiled javascript on the pi - it improves the startup times a decent amount

src/public/js/npm.js
```diff
@@ -0,0 +1,13 @@
+// This file is autogenerated via the `commonjs` Grunt task. You can require() this file in a CommonJS environment.
```
Page-: There's no Gruntfile included that would generate this?

start
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Make sudo actually work
+HOSTNAME=$(cat /etc/hostname)
```
Page-: This can be `HOSTNAME=$(</etc/hostname)` to avoid the need for cat

start
```diff
@@ -0,0 +1,17 @@
...
+echo "127.0.1.1 $HOSTNAME" >> /etc/hosts
+hostname $HOSTNAME
+
+#Set the root password as root if not set as an ENV variable
+export PASSWD=${PASSWD:=root}
+#Set the root password
+echo "root:$PASSWD" | chpasswd
```
Page-: Can also be `chpasswd <<< "root:$PASSWD"` - although it's fine as it is

start
```diff
@@ -0,0 +1,17 @@
...
+
+#cd /app/src && npm install
+cd /app/src && npm start &
```
Page-: It would be preferable to use a relative path so we don't have to rely on being in a particular place on a user's filesystem

pcarranzav:
@Page- I still have to test the last changes, but I _think_ these last commits address all of your comments.

Page-:
Ok, and it's good to ping when they're done because github doesn't notify when a PR gets changed (or at least it doesn't notify for me :(  )

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM resin/armv7hf-node:0.12.2
```
Page-: We have magic cross-platform build vars that would be good to use, @abresas ?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN apt-get update && apt-get install -y \
+	bind9 \
+	bridge-utils \
+	connman \
+	iptables \
+	libdbus-1-dev \
+	libexpat-dev \
+	net-tools \
+	usbutils \
+	wireless-tools
```
Page-: `&& rm -rf /var/lib/apt/lists/*`

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY ./assets/bind /etc/bind
+
+RUN mkdir -p /usr/src/app/
+COPY package.json /usr/src/app/
+RUN cd /usr/src/app && JOBS=MAX npm install --unsafe-perm --production && npm cache clean
+
+COPY . /usr/src/app
+RUN /usr/src/app/node_modules/.bin/coffee -c /usr/src/app/src
+RUN chmod +x /usr/src/app/start
+RUN cd /usr/src/app && /usr/src/app/node_modules/.bin/bower install
```
Page-: You can use caching for bower the same as for `npm` via
```
COPY bower.json /usr/src/app/
RUN cd /usr/src/app && /usr/src/app/node_modules/.bin/bower install
```

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+	libexpat-dev \
+	net-tools \
+	usbutils \
+	wireless-tools	
+
+COPY ./assets/bind /etc/bind
+
+RUN mkdir -p /usr/src/app/
+COPY package.json /usr/src/app/
+RUN cd /usr/src/app && JOBS=MAX npm install --unsafe-perm --production && npm cache clean
```
Page-: If you add a `WORKDIR /usr/src/app` you can avoid the need for all the `cd`s and the like

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+port = process.env.PORT or 8080
+
+server = null
+ssidList = null
+dnsServer = null
+
+getIptablesRules = (callback) ->
+	async.retry {times: 100, interval: 100}, (cb) ->
+		try
+			myIP = os.networkInterfaces().tether[0].address
```
Page-: I like:
```coffee
Promise.try ->
	os.networkInterfaces().tether[0].address
.nodeify(cb)
```
but it's minor

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+			cb(null, myIP)
+		catch err
+			cb(err)
+	, (err, myIP) ->
+		throw err if err?
```
Page-: `throw err`?! This will crash the process, is there any reason for being so extreme?

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+				table: 'nat'
+				rule: "TETHER -p udp --dport 53 -j DNAT --to-destination #{myIP}:53"
+		]
+
+
+startServer = (wifi) ->
+	console.log("Getting networks list")
+	wifi.getNetworksAsync().then (list) ->
+		ssidList = list
+	.then ->
```
Page-: This `.then` can be dropped

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+		wifi.openHotspotAsync(ssid, passphrase)
+	.then ->
+			console.log("Hotspot enabled")
```
Page-: Double indent

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+			dnsServer = spawn('named', ['-f'])
+			getIptablesRules (err, iptablesRules) ->
```
Page-: `err` isn't handled

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+				iptables.appendManyAsync(iptablesRules)
+				.then ->
+					console.log("Captive portal enabled")
+					server = app.listen port, ->
```
Page-: `err` isn't handled

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+	.spread (wifi, properties) ->
+		wifi = Promise.promisifyAll(wifi)
+		console.log("WiFi initialized")
+
+		app.use(bodyParser())
+		app.use(express.static(__dirname + '/public'))
+		app.get '/ssids', (req, res) ->
+			res.send(ssidList)
+		app.post '/connect', (req, res) ->
+			if req.body.ssid and req.body.passphrase
```
Page-: It would be good to do:
```
if not (req.body.ssid and req.body.passphrase)
	return res.sendStatus(400)
...
```
As it currently stands the request will hang if one of those params is missing

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+				console.log("Selected " + req.body.ssid)
+				res.send('OK')
+				server.close()
+				iptables.delete { table: 'nat', rule: "PREROUTING -i tether -j TETHER"}, ->
```
Page-: Is there a possibility of an `err`?

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+					iptables.flush 'nat', 'TETHER', ->
```
Page-: Is there a possibility of an `err`?

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+						dnsServer.kill()
+						console.log("Server closed and captive portal disabled")
+						wifi.joinWithAgentAsync(req.body.ssid, req.body.passphrase)
+						.then ->
+							console.log("Joined! Exiting.")
+							retryCallback()
+						.catch (err) ->
+							console.log(err)
```
Page-: `console.log('Error joining network', err, err.stack)`
- This way we can easily see where the error comes from

src/app.coffee
```diff
@@ -0,0 +1,120 @@
...
+							return startServer(wifi)
+		app.use (req, res) ->
+			res.redirect('/')
+
+		# Create TETHER iptables chain (will silently fail if it already exists)
+		iptables.createChainAsync('nat', 'TETHER')
+		.catch ->
```
Page-: Slightly nicer as
```coffee
.catch -> # Ignore
.then ->
```

Page-: Or maybe somewhere:
```coffee
ignoreError = _.noop
...
.catch(ignoreError)
```

start
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+npm start
+
+while true
+do
+	echo "App exited but container is still running"
```
Page-: ?

src/app.coffee
pcarranzav: In theory, no - if we are here then the rule exists and nothing should go wrong

pcarranzav: But then again, `exec` could fail for random reasons...

src/app.coffee
pcarranzav: Well if this doesn't work, the tether interface is down, which means something is _very_ wrong, and it looks like something we can't recover from... but I'm moving the `throw` to the caller of this function.

start
pcarranzav: This app is meant to be integrated with the user's app (running user code after line 3). I'm leaving the container running so that users who test this can use the Web Terminal after the node process exits.

pcarranzav: (At least until we integrate it to the Host OS)

pcarranzav:
@Page- done, except for the last note, as I think it's wontfix, what do you think?

start
Page-: Ah, can you add " to allow web terminal access" or something

Dockerfile.template
```diff
@@ -0,0 +1,32 @@
...
+COPY ./assets/bind /etc/bind
+
+RUN mkdir -p /usr/src/app/
+WORKDIR /usr/src/app
+
+COPY package.json ./
+RUN JOBS=MAX npm install --unsafe-perm --production && npm cache clean
+
+COPY bower.json .bowerrc ./
+RUN ./node_modules/.bin/bower --allow-root install
```
Page-: `&& bower cache clean` - forgot it before sorry

pcarranzav:
@Page- :+1:

package.json
```diff
@@ -0,0 +1,22 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Pablo Carranza Vélez <pablo@resin.io>",
+  "license": "ISC",
+  "dependencies": {
+    "async": "^1.3.0",
+    "bluebird": "^2.9.34",
+    "body-parser": "^1.13.1",
+    "bower": "^1.4.1",
+    "coffee-script": "^1.9.3",
```
Page-: You should use `~` for coffee-script, they have a habit of making breaking changes between minor versions

Page-:
LGTM

PR #9, opened 2015-08-10
Use file to save connections

pcarranzav:
Connman can be a bit mysterious and not save connections every now and then. So we keep our own list.

Depends on #8

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -5,6 +5,7 @@ app = express()
 bodyParser = require('body-parser')
 iptables = Promise.promisifyAll(require('./iptables'))
 spawn = require('child_process').spawn
+exec = require('child_process').exec
 os = require('os')
 async = require('async')
 
@@ -16,6 +17,11 @@ port = process.env.PORTAL_PORT or config.port
 server = null
 ssidList = null
 dnsServer = null
+
+connectionFile = '/data/connections.json'
+connectionsFromFile = null
+if fs.existsSync(connectionFile)
+	connectionsFromFile = require(connectionFile)
 
 ignore = ->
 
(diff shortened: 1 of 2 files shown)


src/app.coffee
```diff
@@ -58,6 +64,32 @@ startServer = (wifi) ->
...
+
+saveToFile = (ssid, passphrase) ->
+	if connectionsFromFile?
+		connectionsFromFile.push({ ssid, passphrase })
+	else
+		connectionsFromFile = [ { ssid, passphrase } ]
+	return new Promise (resolve, reject) ->
+		fs.writeFile connectionFile, JSON.stringify(connectionsFromFile), (err) ->
+			return reject(err) if err?
+			exec 'sync', (err) ->
```
petrosagg: It's probably better to just `fsync()` the file https://nodejs.org/api/fs.html#fs_fs_fsync_fd_callback

src/app.coffee
```diff
@@ -58,6 +64,32 @@ startServer = (wifi) ->
...
+	else
+		startServer(wifi)
+
+saveToFile = (ssid, passphrase) ->
+	if connectionsFromFile?
+		connectionsFromFile.push({ ssid, passphrase })
+	else
+		connectionsFromFile = [ { ssid, passphrase } ]
+	return new Promise (resolve, reject) ->
+		fs.writeFile connectionFile, JSON.stringify(connectionsFromFile), (err) ->
```
Page-: You can use `fs = Promise.promisifyAll require 'fs'` and then `fs.writeFileAsync`, to avoid needing the manual `new Promise...`

src/app.coffee
```diff
@@ -58,6 +64,32 @@ startServer = (wifi) ->
...
+				console.log('Joined! Exiting.')
+				retryCallback()
+			.catch(ignore)
+		.then ->
+			startServer(wifi)
+	else
+		startServer(wifi)
+
+saveToFile = (ssid, passphrase) ->
+	if connectionsFromFile?
```
Page-: This could just be:
```coffee
connectionsFromFile ?= []
connectionsFromFile.push({ ssid, passphrase })
```

pcarranzav:
@Page- @petrosagg thanks! Done.

src/app.coffee
```diff
@@ -17,6 +19,11 @@ server = null
 ssidList = null
 dnsServer = null
 
+connectionFile = '/data/connections.json'
+connectionsFromFile = null
+if fs.existsSync(connectionFile)
```
Page-: This can just be:
```coffee
connectionsFromFile =
	try
		require(connectionFile)
	catch
		[]
```

Page-: Or you can switch it around to this if you like
```coffee
try
	connectionsFromFile = require(connectionFile)
catch
	connectionsFromFile = []

Comment (pcarranzav) [inline src/app.coffee]: I kept it as null in the catch, as later I do `if connectionsFromFile?`

Comment (Page-) [inline src/app.coffee]: Part of the reason for that was to avoid the need for those later checks :P

File: src/app.coffee [hunk 5]
```diff
@@ -58,6 +66,31 @@ startServer = (wifi) ->
 					throw err if err?
 					console.log('Server listening')
 
+connectOrStartServer = (wifi, retryCallback) ->
+	if connectionsFromFile?
```
Page-: Can you use:
```coffee
Promise.try ->
	return if connectionsFromFile?
	console...
.then ->
	startServer(wifi)
```
so that we always use the same `.then -> startServer` (maybe `.finally` is best though?)

pcarranzav:
@Page- @petrosagg what do you think?

Page-:
LGTM

PR #12, opened 2015-08-18
Don't crash with no networks

pcarranzav:
Currently, as connman-simplified returns an error, when the scan results in no networks found we crash.
This PR fixes this by catching the error, and showing an error message on the web UI.

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -51,7 +51,10 @@ getIptablesRules = (callback) ->
 
 startServer = (wifi) ->
 	console.log('Getting networks list')
-	wifi.getNetworksAsync().then (list) ->
+	wifi.getNetworksAsync()
+	.catch (err) ->
+		throw err unless err.message == 'No WiFi networks found'
+	.then (list) ->
 		ssidList = list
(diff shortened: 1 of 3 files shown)


lorenzo-stoakes:
LGTM

src/app.coffee
```diff
@@ -51,7 +51,11 @@ getIptablesRules = (callback) ->
 
 startServer = (wifi) ->
 	console.log('Getting networks list')
-	wifi.getNetworksAsync().then (list) ->
+	wifi.getNetworksAsync()
+	.catch (err) ->
+		throw err unless err.message == 'No WiFi networks found'
```
Page-: It makes me so sad they don't extend error with an `err.code`, meaning we have to do a string check which could change at any point :(

Page-:
LGTM

src/app.coffee
pcarranzav: Yeah, comparing strings is not very nice... I've already contributed to connman-simplified, so I might add this on a future PR...

PR #13, opened 2015-08-18
Fix saveFile + closeHotspot at startup

pcarranzav:
- Fix saveFile (writeAsync was being called with wrong arguments).
- Ensure Hotspot is closed on startup (otherwise program breaks after an update without reboot).

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -85,7 +85,8 @@ saveToFile = (ssid, passphrase) ->
 	connectionsFromFile.push({ ssid, passphrase })
 	fs.openAsync(connectionFile, 'w')
 	.tap (fd) ->
-		fs.writeAsync(fd, JSON.stringify(connectionsFromFile))
+		buf = new Buffer(JSON.stringify(connectionsFromFile))
+		fs.writeAsync(fd, buf, 0, buf.length, null)
 	.tap (fd) ->
 		fs.fsyncAsync(fd)
 	.then (fd) ->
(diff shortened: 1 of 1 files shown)


src/app.coffee
```diff
@@ -85,7 +85,8 @@ saveToFile = (ssid, passphrase) ->
 	connectionsFromFile.push({ ssid, passphrase })
 	fs.openAsync(connectionFile, 'w')
 	.tap (fd) ->
-		fs.writeAsync(fd, JSON.stringify(connectionsFromFile))
+		buf = new Buffer(JSON.stringify(connectionsFromFile))
```
petrosagg: Why do you use a buffer here?

pcarranzav: We're using node 0.10.38, and it either expects a buffer or to specify encoding (see https://github.com/joyent/node/blob/v0.10.38-release/lib/fs.js#L497) - I felt this was cleaner.

Page-: https://nodejs.org/docs/v0.10.0/api/fs.html#fs_fs_writefile_filename_data_options_callback says that the default is 'utf8', so just a string should work :S

pcarranzav: That's for fs.writeFile, we're using [fs.write](https://nodejs.org/docs/v0.10.0/api/fs.html#fs_fs_write_fd_buffer_offset_length_position_callback) so that then we do fs.fsync.

petrosagg:
SGTM

src/app.coffee
Page-: Ah, I clicked `fs.write` in the contents, but it's linked to the wrong place..

PR #15, opened 2015-08-27
add responsive meta tag and load style first

craig-mulligan:
That meta tag should make it scale better on mobile.


pcarranzav:
@craig-mulligan LGTM

PR #28, opened 2016-02-03
Disable connman service and pass correct dbus socket for when systemd…

### LINKED ISSUE #26: Need environment variable to work with systemd
When systemd is enabled (`ENV INITSYSTEM on`), the dbus socket will be overwritten by the container's dbus. We need to start the node app with `DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket`. We also need to `RUN systemctl disable connman` in Dockerfile.template.

pcarranzav:
… is enabled.
Should fix #26

### DIFF
--- a/start
+++ b/start
@@ -1,5 +1,9 @@
 #!/bin/bash
-npm start
+if [[ "$INITSYSTEM" == "on" ]]; then
+	DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket npm start
+else
+	npm start
+fi
 
 while true
 do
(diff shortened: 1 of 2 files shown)


lorenzo-stoakes:
LGTM, pending tests!

start
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
-npm start
+if [[ "$INITSYSTEM" == "on" ]]; then
+	DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket npm start
```
Page-: I think this can be `export DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket` and then just one `npm start` below

pcarranzav: Yeah, but I wanted to keep this env var assignment exclusive to the node app.

Page-: Ok

PR #29, opened 2016-02-20
Fix typo in config key

andrewn:
Fixes issue where passphrase set via config file was never picked up by application


Page-:
Thanks :)

PR #39, opened 2016-06-15
Use the more generic 0.10 tag on the Dockerfile

### LINKED ISSUE #38: Base image does not exist for raspberrypi3
`FROM resin/%%RESIN_MACHINE_NAME%%-node:0.10.40`

petrosagg:
Using this tag avoids issues with non-existent device images. This can
happen if a new devices gets support in resin and at the time node
version has advanced, so old docker images don't exist.

Fixes #38

Signed-off-by: Petros Angelatos <petrosagg@gmail.com>

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,4 +1,4 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-node:0.10.40
+FROM resin/%%RESIN_MACHINE_NAME%%-node:0.10
 
 RUN apt-get update && apt-get install -y \
 	bind9 \


nghiant2710:
SGTM :+1:

alexandrosm:
The correct solution seems to be that we maintain a minimum version we
build for all boards, and wifi-connect (and all our other projects btw) is
at that version or later. How do we decide the minimum version.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

PR #44, opened 2016-07-22
Always use the /host_run dbus socket

pcarranzav:
@petrosagg

### DIFF
--- a/start
+++ b/start
@@ -1,9 +1,6 @@
 #!/bin/bash
-if [[ "$INITSYSTEM" == "on" ]]; then
-	DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket npm start
-else
-	npm start
-fi
+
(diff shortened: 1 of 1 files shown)


petrosagg:
SGTM

start
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
-if [[ "$INITSYSTEM" == "on" ]]; then
-	DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket npm start
-else
-	npm start
-fi
+
+DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket npm start
```
Page-: Wouldn't it be better for this to use `/host/run`?

petrosagg: Pablo said old supervisors don't bind-mount `/host` so until we update everything we have to use this

PR #49, opened 2016-10-06
Network manager support

josephroberts:


src/systemd.coffee
```diff
@@ -17,3 +17,17 @@ exports.stop = (unit, mode='fail') ->
 	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
 	.then (manager) ->
 		manager.StopUnitAsync(unit, mode)
+
+exports.exists = (unit, mode='fail') ->
+	new Promise (resolve, reject) ->
```
Page-: There's no need to create a promise here, you can just return the promise chain directly, see http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern

src/utils.coffee
```diff
@@ -15,3 +15,12 @@ exports.durableWriteFile = (file, data) ->
...
 	.then(fs.closeAsync)
+
+exports.copyFile = (source, target) ->
+	fs.readFileAsync(source)
+	.then (rf) ->
+		fs.writeFileAsync(target, rf)
+		
+exports.exists = (path) ->
+	new Promise (resolve) ->
+        fs.exists(path, resolve)
```
Page-: You can just use `fs.existsAsync(path)` directly

src/networkManager.coffee
```diff
@@ -0,0 +1,108 @@
...
+				bus.getInterfaceAsync(SERVICE, connections[index], TECHNOLOGY_INTERFACE + ".Connection")
+				.then (connection) ->
+					console.log("delete")
+					connection.DeleteAsync()
+				.then ->
+					index += 1
+					console.log(index)
+					if index == connections.length
+						console.log("jhjhj")
+						resolve
```
Page-: This doesn't call `resolve`, to call a function with no params you need the brackets

src/networkManager.coffee
```diff
@@ -0,0 +1,108 @@
...
+
+			"""
+
+	console.log('Saving credentials')
+	console.log(data)
+
+	return utils.durableWriteFile(config.persistentConfig, data)
+
+exports.clearCredentials = () ->
+	new Promise (resolve, reject) ->
```
Page-: There's no need to create a promise here, you can just return the promise chain directly, see http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern

src/networkManager.coffee
```diff
@@ -0,0 +1,108 @@
...
+DEFAULT_CONNECTIONS = 2
+
+exports.start = () ->
+	systemd.start('NetworkManager.service')
+
+exports.stop = () -> 
+	systemd.stop('NetworkManager.service')
+
+exports.isSetup = () ->
+	new Promise (resolve, reject) ->
```
Page-: There's no need to create a promise here, you can just return the promise chain directly, see http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern

src/connman.coffee
```diff
@@ -5,13 +5,57 @@ dbus = new DBus()
...
-exports.waitForConnection = (timeout) ->
-	console.log('Waiting for connman to connect..')
+exports.start = ->
+	systemd.start('connman.service')
+
+exports.stop = ->
+	systemd.stop('connman.service')
+
+exports.isSetup = ->
+	new Promise (resolve, reject) ->
```
Page-: There's no need to create a promise here, you can just return the promise chain directly, see http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern

src/app.coffee
```diff
@@ -25,45 +22,81 @@ app.post '/connect', (req, res) ->
...
+	if result
+		console.log('Using NetworkManager.service')
+		manager = networkManager
+	else
+		console.log('Using connman.service')
+		manager = connman
+.then ->
+	wifiScan.scanAsync()
+.then (results) ->
+	ssids = results
```
Page-: This line doesn't appear to actually do anything?

josephroberts: Surely it stores the `ssids` so they are available to other parts of the app?

src/utils.coffee
josephroberts: Like this? Does the function return the last line explicitly i.e. without writing return?

src/systemd.coffee
josephroberts: Like this? Does the function return the last line explicitly i.e. without writing return?

josephroberts: Like this? Does the function return the last line explicitly i.e. without writing return?

src/utils.coffee
josephroberts: Like this? Does the function return the last line explicitly i.e. without writing return?

src/connman.coffee
josephroberts: Like this? Does the function return the last line explicitly i.e. without writing return?

src/app.coffee
Page-: It stores it in a local variable, which instantly falls out of scope, at least as far as I can tell that's what it was doing

src/systemd.coffee
Page-: Yes it does - looks good now

Page-:
Can you also rebase this to latest master and use `npm run lint` to check for linting errors please

src/app.coffee
josephroberts: it puts it in a member variable here https://github.com/resin-io/resin-wifi-connect/blob/a464c3bf35ef640d69d693962d23b9e44557d4d5/src/app.coffee#L16

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+
+exports.start = ->
+	systemd.start('NetworkManager.service')
+
+exports.stop = ->
+	systemd.stop('NetworkManager.service')
+
+exports.isSetup = ->
+	getConnections()
+	.then (connections) ->
```
Page-: This should be
```coffee
.map(validateConnection)
.then (results) ->
	return true in results

File: src/connman.coffee [hunk 9]
```diff
@@ -5,13 +5,54 @@ dbus = new DBus()
...
+		if exists
+			utils.copyFile(config.persistentConfig, config.connmanConfig)
+			.then ->
+				return true
+		else
+			return false
+
+exports.setCredentials = (ssid, passphrase) ->
+	data = """
+				[service_home_ethernet]
```
Page-: This block should lose two levels of indentation

src/connman.coffee
```diff
@@ -5,13 +5,54 @@ dbus = new DBus()
...
+
+exports.stop = ->
+	systemd.stop('connman.service')
+
+exports.isSetup = ->
+	fs.existsAsync(config.persistentConfig)
+	.then (exists) ->
+		if exists
+			utils.copyFile(config.persistentConfig, config.connmanConfig)
+			.then ->
```
Page-: This is nicer as:
```coffee
.return(true)
```

src/app.coffee
```diff
@@ -16,54 +13,97 @@ app.use(bodyParser.json())
...
+			console.log('Credentials not found')
+			hotspot.start(manager)
+			.catch (e) ->
+				console.log(e)
+				console.log('Exiting')
+				process.exit()
 
 app.listen(80)
+
+retry = null
```
Page-: There's no need to declare this here, it will automatically declared by the statements below

src/app.coffee
```diff
@@ -16,54 +13,97 @@ app.use(bodyParser.json())
...
+if process.argv[2] == '--retry=true'
+	console.log('Retry enabled')
+	retry = true
+else if process.argv[2] == '--retry=false'
+	console.log('Retry disabled')
+	retry = false
+else if typeof(process.argv[2]) == 'undefined'
```
Page-: This would be nicer as
```coffee
else if not process.argv[2]?
```
- the `?` existential operator is what we normally use to check for null/undefined

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+		Promise.all(buffer)
+		.then (result) ->
+			bus.getInterfaceAsync(SERVICE, '/org/freedesktop/NetworkManager', 'org.freedesktop.NetworkManager')
+			.then (manager) ->
+				_manager = manager
+				_manager.ActivateConnectionAsync('/', devices[result.indexOf(true)], '/')
+			.then ->
+				new Promise (resolve, reject, onCancel) ->
+					handler = (value) ->
+						if value == 70
```
Page-: It would be nice to declare this as a var higher up in the module, rather than having a magic value (which has no meaning by itself), ie
```coffee
MEANINGFUL_NAME_FOR_VALUE = 70
```

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+							_manager.removeListener('StateChanged', handler)
+							resolve()
+
+					_manager.on('StateChanged', handler)
+
+					_manager.CheckConnectivityAsync()
+					.then (state) ->
+						if state == 5
```
Page-: It would be nice to declare this as a var higher up in the module, rather than having a magic value (which has no meaning by itself), ie
```coffee
MEANINGFUL_NAME_FOR_STATE = 5
```

src/app.coffee
Page-: When I searched the file with chrome it didn't find any other instance of `ssids` :S  It's ok in that case

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+				'key-mgmt': 'wpa-psk',
+				'psk': passphrase,
+			}
+		}
+
+		manager.AddConnectionAsync(connection)
+
+exports.clearCredentials = ->
+	getConnections()
+	.then (connections) ->
```
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+		for connection in connections
+			buffer.push(deleteConnection(connection))
+
+		Promise.all(buffer)
+
+exports.connect  = (timeout) ->
+	getDevices()
+	.then (devices) ->
+		buffer = []
+		for device in devices
```
josephroberts: done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+
+		_manager = null
+		Promise.all(buffer)
+		.then (result) ->
+			bus.getInterfaceAsync(SERVICE, '/org/freedesktop/NetworkManager', 'org.freedesktop.NetworkManager')
+			.then (manager) ->
+				_manager = manager
+				_manager.ActivateConnectionAsync('/', devices[result.indexOf(true)], '/')
+			.then ->
+				new Promise (resolve, reject, onCancel) ->
```
josephroberts: Done, no linter warning though?

src/networkManager.coffee
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+							resolve()
+
+					setTimeout ->
+						_manager.removeListener('StateChanged', handler)
+						reject()
+					, timeout
+
+getConnections = ->
+	bus.getInterfaceAsync(SERVICE, '/org/freedesktop/NetworkManager/Settings', 'org.freedesktop.NetworkManager.Settings')
+	.then (settings) ->
```
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+		settings.ListConnectionsAsync()
+
+getConnection = (connection) ->
+	bus.getInterfaceAsync(SERVICE, connection, 'org.freedesktop.NetworkManager.Settings.Connection')
+
+validateConnection = (connection) ->
+	getConnection(connection)
+	.then (connection) ->
```
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+		connection.GetSettingsAsync()
+		.then (settings) ->
+			return settings.connection.id in WHITE_LIST
+
+deleteConnection = (connection) ->
+	_connection = null
+	getConnection(connection)
+	.then (connection) ->
+		_connection = connection
```
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+	.then (manager) ->
+		manager.GetDevicesAsync()
+
+getDevice = (device) ->
+	bus.getInterfaceAsync(SERVICE, device, 'org.freedesktop.NetworkManager.Device')
+
+validateDevice = (device, type) ->
+	getDevice(device)
+	.then (device) ->
+		device.getPropertyAsync('DeviceType')
```
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,141 @@
...
+		.then (property) ->
+			return property == type
+
+unpack = (str) ->
```
josephroberts: Done

josephroberts: To convert the SSID from string to bytes, removed now

src/systemd.coffee
```diff
@@ -17,3 +17,13 @@ exports.stop = (unit, mode = 'fail') ->
 	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
 	.then (manager) ->
 		manager.StopUnitAsync(unit, mode)
+
+exports.exists = (unit, mode = 'fail') ->
+	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
+	.then (manager) ->
```
josephroberts: Done

src/connman.coffee
josephroberts: Done

src/connman.coffee
josephroberts: Done

src/app.coffee
josephroberts: Done

src/app.coffee
josephroberts: Done

josephroberts:
Pending completion of #52 
I don't feel any other issues need resolving as the NM feature set will match the connman feature set once #52 is complete. @petrosagg would you agree?

Are you happy with the tests listed in ManualTests.md?

src/networkManager.coffee
Page-: Ah, I didn't see it was used like that, in that case a better alternative is:
```coffee
.filter (device) ->
	validateDevice(device, 2)
.then (validDevices) ->
	if validDevices.length is 0
		throw new Error('No valid devices to connect')
	deviceToConnect = validDevices[0] # Should we be using the others?
```

package.json
```diff
@@ -4,7 +4,6 @@
   "repository": "https://github.com/pcarranzav/resin-wifi-connect",
   "description": "A tool to allow wifi configuration to be set via a captive portal",
   "scripts": {
-    "start": "cd src && node app.js",
```
Page-: I think it'd be `npm start -- --restart=true`

josephroberts: Yeah that works :)

src/networkManager.coffee
josephroberts: Done

src/networkManager.coffee
```diff
@@ -0,0 +1,122 @@
...
+	getConnections()
+	.map(deleteConnection)
+
+exports.connect  = (timeout) ->
+	getDevices()
+	.filter (device) ->
+		validateDevice(device)
+	.then (validDevices) ->
+		if validDevices.length is 0
+			reject()
```
Page-: This should be a throw, `reject` is undefined here (also because there's no return it'd still try to run the code below regardless)

src/networkManager.coffee
```diff
@@ -0,0 +1,122 @@
...
+	.then (settings) ->
+		settings.AddConnectionAsync(connection)
+
+exports.clearCredentials = ->
+	getConnections()
+	.map(deleteConnection)
+
+exports.connect  = (timeout) ->
+	getDevices()
+	.filter (device) ->
```
Page-: This can be `.filter(validateDevice)`

alexandrosm:
@josephroberts @Page- @petrosagg -- where are we with this PR?

josephroberts:
@alexandrosm pending completion of #52, testing and any comments from pepe. I will get it done this week.

src/networkManager.coffee
josephroberts: Done

josephroberts:
#52 Tested on `WPA-PSK` and `WPA2-PSK`. Not tested on `WEP`, `Dynamic WEP`, `Ad-Hoc WPA-PSK` or `WPA-Enterprise` because we don't have a device capable of creating these types of AP.

Do we need to support any of the un-tested types @petrosagg ?
Any other comments @petrosagg ?

I will test thoroughly once your happy :)

src/networkManager.coffee
```diff
@@ -0,0 +1,125 @@
...
+	systemd.start('NetworkManager.service')
+
+exports.stop = ->
+	systemd.stop('NetworkManager.service')
+
+exports.isSetup = ->
+	getConnections()
+	.map(validateConnection)
+	.then (results) ->
+		return false in results
```
Page-: This returns true if any connection failed to validate, is that correct?

src/networkManager.coffee
josephroberts: @Page- I couldn't figure this out. Could you point me to the docs on `filter`, `map` etc. please. My google fu is failing me

src/networkManager.coffee
josephroberts: Yes, but it is confusing. I will swap it over now.

shaunmulligan:
@petrosagg and @josephroberts can we get this merged this week?

josephroberts:
@shaunmulligan yep gonna test fully today. I would really like it if someone else could also do the same because I am very aware we could leave devices without internet if something goes wrong

src/networkManager.coffee
Page-: http://bluebirdjs.com/docs/api/filter.html is the docs for bluebird filter (although if you search for `mdn filter` it'll give you the general js docs - the format is the same for both though).

Really though all this change would be doing is to have the arguments passed directly to `validateDevice` and avoid creating the anonymous function in the middle

josephroberts:
I have released under the tag `2.0.0` and not merged into master. This will help to prevent people from accidentally pulling the master branch in their docker file and breaking any existing code. This is necessary because the control flow and command to start the app have changed.

PR #57, opened 2016-11-16
Network manager support

josephroberts:
@petrosagg @Page- @jviotti @alexandrosm please review. 

It works in the same way as the previous version if the user calls either; `./start` or `node app.js`, it breaks if the user has integrated the start script into their own code.

Passing all tests inside `ManualTests.md`

I think we can merge into master with the changes made but it should still be released as a new version `2.0.0`


src/app.coffee
```diff
@@ -18,52 +15,107 @@ app.use(express.static(__dirname + '/public'))
...
-	]
-	# XXX: make it so this delay isn't needed
-	.delay(1000)
-	.then ->
-		connman.waitForConnection(15000)
+	hotspot.stop(manager)
 	.then ->
-		utils.durableWriteFile(config.persistentConfig, data)
+		manager.setCredentials(req.body.ssid, req.body.passphrase)
 	.then ->
```
jviotti: This can be `.then(run)`.

src/app.coffee
```diff
@@ -18,52 +15,107 @@ app.use(express.static(__dirname + '/public'))
 
 ssids = []
 
+error = (e) ->
+	console.log(e)
+	if retry
+		console.log('Retrying')
+		console.log('Clearing credentials')
+		manager.clearCredentials()
+		.then ->
```
josephroberts: Done

src/app.coffee
```diff
@@ -18,52 +15,107 @@ app.use(express.static(__dirname + '/public'))
...
+			run()
+		.catch (e) ->
```
josephroberts: Done

src/app.coffee
josephroberts: Done

src/app.coffee
```diff
@@ -18,52 +15,107 @@ app.use(express.static(__dirname + '/public'))
...
-	.then ->
-		connman.waitForConnection(15000)
+	hotspot.stop(manager)
 	.then ->
-		utils.durableWriteFile(config.persistentConfig, data)
+		manager.setCredentials(req.body.ssid, req.body.passphrase)
 	.then ->
-		process.exit()
+		run()
 	.catch (e) ->
```
josephroberts: Done

josephroberts:
@petrosagg @Page- @jviotti Please can I get two LGTMs' so I can merge this. Thanks

src/systemd.coffee
```diff
@@ -17,3 +15,12 @@ exports.stop = (unit, mode = 'fail') ->
 	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
 	.then (manager) ->
 		manager.StopUnitAsync(unit, mode)
+
+exports.exists = (unit, mode = 'fail') ->
+	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
+	.call('ListUnitsAsync')
+	.then (units) ->
+		for obj in units
```
jviotti: Could this be replaced with `_.includes`?

src/utils.coffee
```diff
@@ -15,3 +15,8 @@ exports.durableWriteFile = (file, data) ->
 		fs.openAsync(path.dirname(file), 'r', constants.O_DIRECTORY)
 	.tap(fs.fsyncAsync)
 	.then(fs.closeAsync)
+
+exports.copyFile = (source, target) ->
```
jviotti: What is the size of the files meant to be copied here? If this only applies to small files then its fine, however since the whole thing gets read into memory, we could switch to a stream based copy approach.

You can always use one of the many "copy" modules in NPM which implement stream-based copying to be on the safe side (mainly for embedded devices).

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+	getDevices()
+	.filter(validateDevice)
+	.then (validDevices) ->
+		if validDevices.length is 0
+			throw ('No valid devices found.')
+		getConnections()
+		.filter(validateConnection)
+		.then (validConnections) ->
+			if validConnections.length is 0
+				throw ('No valid connections found.')
```
jviotti: We shouldn't be throwing strings, but `Error` instances, since a string is not an error object, and thus lacks any useful debugging information:

```js
throw new Error('No valid connections found.')
```

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+exports.clearCredentials = ->
+	getConnections()
+	.map(deleteConnection)
+
+exports.connect  = (timeout) ->
+	getDevices()
+	.filter(validateDevice)
+	.then (validDevices) ->
+		if validDevices.length is 0
+			throw ('No valid devices found.')
```
jviotti: Same as below, lets throw an `Error`.

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+
+						# But try to read in case we registered the event handler
+						# after is was already connected
+						manager.CheckConnectivityAsync()
+						.then (state) ->
+							if state == NM_CONNECTIVITY_FULL or state == NM_CONNECTIVITY_LIMITED
+								manager.removeListener('StateChanged', handler)
+								resolve()
+
+						setTimeout ->
```
jviotti: I'm not sure if this is the intended behaviour, but `setTimeout` will *always* be called right after we attempt to check the connectivity above. This means that `setTimeout` will start when `manager.CheckConnectivityAsync()` still didn't resolve, and if such function takes more than `timeout` time resolve for whatever reason, the whole thing will be rejected as a consequence of the timeout handler.

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+							manager.removeListener('StateChanged', handler)
```
jviotti: If this code portion is called, then we attempt to remove the listener, which might resolve the promise, but then we reject it right afterwards. This means that if `handler` is called with `NM_STATE_CONNECTED_GLOBAL`, the promise gets rejected right away, but then attempts to get resolved, which as far as I remember, will cause Bluebird to throw an error saying that the promise has already been rejected.

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+
+						# Listen for 'Connected' signals
+						manager.on('StateChanged', handler)
+
+						# But try to read in case we registered the event handler
+						# after is was already connected
+						manager.CheckConnectivityAsync()
+						.then (state) ->
+							if state == NM_CONNECTIVITY_FULL or state == NM_CONNECTIVITY_LIMITED
+								manager.removeListener('StateChanged', handler)
```
jviotti: Same thing as below. `handler` resolves if `value` is  `NM_STATE_CONNECTED_GLOBAL`, but we also resolve right afterwards. Attempting to resolve twice might cause Bluebird issues.

jviotti: Also, `handler` might attempt to remove the `StateChanged` listener, which is what we're already doing here.

src/connman.coffee
```diff
@@ -1,37 +1,76 @@
...
+	"""
+
+	console.log('Saving connection')
+	console.log(connection)
+
+	utils.durableWriteFile(config.persistentConfig, connection)
+
+exports.clearCredentials = ->
+	fs.unlinkAsync(config.persistentConfig)
+	.catch code: 'ENOENT', (e) ->
```
jviotti: What about using `_.noop` here?

```js
.catch(code: 'ENOENT', _.noop)
```

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+
+deleteConnection = (connection) ->
+	getConnection(connection)
+	.then (connection) ->
+		connection.GetSettingsAsync()
+		.then (settings) ->
+			if settings.connection.id not in WHITE_LIST
+				connection.DeleteAsync()
+
+validateConnection = (connection) ->
```
jviotti: This function returns a boolean value representing whether the connection id is white listed or not. If it is, it will return `false`. Maybe we can name the function differently to make the intention clearer? In this case, it could be something like `isConnectionInvalid()`.

src/networkManager.coffee
```diff
@@ -0,0 +1,129 @@
...
+			throw ('No valid devices found.')
+		getConnections()
+		.filter(validateConnection)
+		.then (validConnections) ->
+			if validConnections.length is 0
+				throw ('No valid connections found.')
+			bus.getInterfaceAsync(SERVICE, '/org/freedesktop/NetworkManager', 'org.freedesktop.NetworkManager')
+			.delay(1000) # Delay needed to avoid "Error: org.freedesktop.NetworkManager.UnknownConnection at Error (native)" when activating the connection
+			.then (manager) ->
+				manager.ActivateConnectionAsync(validConnections[0], validDevices[0], '/')
```
jviotti: Why do we only activate the first valid connection and first valid device? What about the rest? Should we be iterating instead?

src/systemd.coffee
josephroberts: Done

src/utils.coffee
josephroberts: it copies a very small txt file so its probably OK?
```
		[service_home_ethernet]
		Type = ethernet
		Nameservers = 8.8.8.8,8.8.4.4

		[service_home_wifi]
		Type = wifi
		Name = #{ssid}
		Passphrase = #{passphrase}
		Nameservers = 8.8.8.8,8.8.4.4

```

src/networkManager.coffee
josephroberts: Done

src/networkManager.coffee
josephroberts: Done

src/networkManager.coffee
josephroberts: I think this is the expected behaviour

src/connman.coffee
```diff
@@ -1,37 +1,76 @@
...
 				if Connected
-					wifi.removeListener('PropertyChanged', handler)
+					manager.removeListener('PropertyChanged', handler)
 					resolve()
 
 			setTimeout ->
-				wifi.removeListener('PropertyChanged', handler)
-				reject()
+				manager.removeListener('PropertyChanged', handler)
+				reject('Timed out')
```
josephroberts: I think this is the expected behaviour

src/networkManager.coffee
josephroberts: Can you explain a bit more please, are you talking about what would happen if the handler and timeout occurred at the exact-ish same time?

src/networkManager.coffee
josephroberts: If one branch of the promise resolves that means the whole promise on line 75 resolves which stops any other branch from resolving. Is this not how it works @jviotti ?

src/connman.coffee
josephroberts: Done

src/networkManager.coffee
josephroberts: resin-wifi-connect supports a single set of credentials with connman, to maintain backward compatibility/same behaviour resin-wifi-connect also supports a single set of credentials

josephroberts:
@Page- would you mind giving a final once over and LGTM so I can get this merged? Thanks

src/systemd.coffee
Page-: It could be replaced with
```coffee
_.some(units, unit)
````
if the value of `unit` key is truthy, but I don't think it makes what is happening clearer

File: src/networkManager.coffee [hunk 16]
```diff
@@ -0,0 +1,129 @@
...
+						# after is was already connected
+						manager.CheckConnectivityAsync()
+						.then (state) ->
+							if state == NM_CONNECTIVITY_FULL or state == NM_CONNECTIVITY_LIMITED
+								manager.removeListener('StateChanged', handler)
+								resolve()
+
+						setTimeout ->
+							manager.removeListener('StateChanged', handler)
+							reject('Timed out')
```
Page-: This should reject with an error, ie `reject(new Error('Timed out'))`

src/hotspot.coffee
```diff
@@ -30,15 +28,18 @@ exports.start = ->
...
 	started = false
 
 	Promise.all [
 		hostapd.stop()
 		dnsmasq.stop()
 	]
 	.then ->
-		systemd.start('connman.service')
+		manager.start()
+	.delay(2000) # Delay needed to give service time to start
```
Page-: Is there a way to wait for the service to start, rather than an arbitrary delay (which may vary from device to device, and how busy it is at the time)

src/connman.coffee
```diff
@@ -1,37 +1,76 @@
...
```
Page-: This should reject with an error, ie `reject(new Error('Timed out'))`

src/connman.coffee
```diff
@@ -1,37 +1,76 @@
...
+
+exports.stop = ->
+	systemd.stop('connman.service')
+
+exports.isSetup = ->
+	fs.statAsync(config.persistentConfig)
+	.then ->
+		utils.copyFile(config.persistentConfig, config.connmanConfig)
+		.return(true)
+	.catch (e) ->
```
Page-: This can be
```coffee
.then ->
	utils.copyFile(config.persistentConfig, config.connmanConfig)
.return(true)
.catchReturn(false)
```

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
+.then ->
+	if clear
+		console.log('Clearing credentials')
+		manager.clearCredentials()
+.then ->
+	manager.isSetup()
+	.then (setup) ->
+		if setup
+			retry = false
+.then (run)
```
Page-: Can you make this `.then(run)`, `.then (run)` means something else in coffeescript, for example the following are equivalent:
```coffee
.then (run), (x)
.then(run, x)
```
but
```coffee
.then (run, x)
```
will fail to compile

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
+			.catch (error)
+		else
+			console.log('Credentials not found')
+			hotspot.stop(manager)
+			.then ->
+				wifiScan.scanAsync()
+			.then (results) ->
+				ssids = results
+				hotspot.start(manager)
+			.catch (error)
```
Page-: Can you make this `.catch(error)` please - explanation is in another comment

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
+			console.log('Credentials found')
+			hotspot.stop(manager)
+			.then ->
+				console.log('Connecting')
+				manager.connect(15000) # Delay needed to allow manager to connect
+			.then ->
+				console.log('Connected')
+				console.log('Exiting')
+				process.exit()
+			.catch (error)
```
Page-: Can you make this `.catch(error)` please - explanation is in another comment

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
-	.then ->
-		utils.durableWriteFile(config.persistentConfig, data)
+	hotspot.stop(manager)
 	.then ->
-		process.exit()
-	.catch (e) ->
-		hotspot.start()
+		manager.setCredentials(req.body.ssid, req.body.passphrase)
+	.then (run)
+	.catch (error)
```
Page-: Can you make this `.catch(error)` please - explanation is in another comment

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
-		connman.waitForConnection(15000)
-	.then ->
-		utils.durableWriteFile(config.persistentConfig, data)
+	hotspot.stop(manager)
 	.then ->
-		process.exit()
-	.catch (e) ->
-		hotspot.start()
+		manager.setCredentials(req.body.ssid, req.body.passphrase)
+	.then (run)
```
Page-: Can you make this `.then(run)` please - explanation is in another comment

src/app.coffee
```diff
@@ -18,52 +15,100 @@ app.use(express.static(__dirname + '/public'))
...
 ssids = []
 
+error = (e) ->
+	console.log(e)
+	if retry
+		console.log('Retrying')
+		console.log('Clearing credentials')
+		manager.clearCredentials()
+		.then (run)
+		.catch (error)
```
Page-: Can you make these `.catch(error)` and `.then(run)` please - explanation is in another comment

src/app.coffee
josephroberts: Done

src/app.coffee
josephroberts: Done

src/app.coffee
josephroberts: Done

src/app.coffee
josephroberts: Done

src/connman.coffee
josephroberts: Done

src/connman.coffee
josephroberts: Done

src/networkManager.coffee
josephroberts: Done

src/systemd.coffee
josephroberts: Done

src/systemd.coffee
```diff
@@ -1,19 +1,63 @@
...
+	.then (objectPath) ->
+		bus.getInterfaceAsync(SERVICE, objectPath, UNIT_INTERFACE)
+	.then (unit) ->
+		unit.getPropertyAsync('ActiveState')
+
+promiseWhile = (condition, action) ->
+	resolver = Promise.defer()
+
+	l = ->
+		if !condition()
```
Page-: If `condition` were to throw then it would crash the nodejs process

src/systemd.coffee
```diff
@@ -1,19 +1,63 @@
...
+			return resolver.resolve()
+		Promise.cast(action()).then(l).catch resolver.reject
```
Page-: If `action` were to throw then it would crash the nodejs process

src/systemd.coffee
```diff
@@ -1,19 +1,63 @@
...
+	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
+	.then (manager) ->
+		manager.GetUnitAsync(unit)
+	.then (objectPath) ->
+		bus.getInterfaceAsync(SERVICE, objectPath, UNIT_INTERFACE)
+	.then (unit) ->
+		unit.getPropertyAsync('ActiveState')
+
+promiseWhile = (condition, action) ->
+	resolver = Promise.defer()
```
Page-: I think this function can be simplified and made safer as:
```coffee
repeat = ->
	Promise.try(condition)
	.then (bool) ->
		return if not bool
		Promise.try(action)
		.then(repeat)
return repeat()

File: src/systemd.coffee [hunk 29]
```diff
@@ -1,19 +1,63 @@
...
+
+	promiseWhile((->
+		currentState != targetState
+	), ->
+		getState(unit)
+		.then (state) ->
+			currentState = state
+		.delay(1000)	
+	).then ->
+		return
```
Page-: What is this for? It can be simplified as `.return(undefined)` though

src/systemd.coffee
```diff
@@ -1,19 +1,63 @@
...
+exports.exists = (unit, mode = 'fail') ->
+	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
+	.call('ListUnitsAsync')
+	.then (units) ->
+		_.has(units[0], unit)
+
+waitUntilState = (unit, targetState) ->
+	currentState = null
+
+	promiseWhile((->
```
Page-: I think this can be formatted slightly nicer as
```coffee
promiseWhile(
	-> currentState != targetState
	->
		getState(unit)
		.then (state) ->
			currentState = state
		.delay(1000)
)
```
although it is definitely a subjective thing

src/systemd.coffee
```diff
@@ -1,19 +1,63 @@
...
 	.then (manager) ->
 		manager.StopUnitAsync(unit, mode)
+	.then ->
+		waitUntilState(unit, 'inactive') 
+
+exports.exists = (unit, mode = 'fail') ->
+	bus.getInterfaceAsync(SERVICE, MANAGER_OBJECT, MANAGER_INTERFACE)
+	.call('ListUnitsAsync')
+	.then (units) ->
+		_.has(units[0], unit)
```
Page-: We only check the first unit?

josephroberts: 'units' is an array containing a single value which is an object. That object contains multiple properties. We are checking for the existence of a property with a key equal to `unit`

src/systemd.coffee
josephroberts: Done

src/systemd.coffee
josephroberts: Done

src/wifi-scan.coffee
```diff
@@ -35,3 +34,6 @@ exports.scanAsync = ->
 
 		# sort by signal strength
 		return networks.sort((a, b) -> b.signal - a.signal)
+	.catch code: 240, (e) ->
+		Promise.delay(1000) # Delay needed to avoid `Command failed: iw wlan0 scan ap-force command failed: Device or resource busy (-16)`
```
josephroberts: I think this delay is fine as it just means that 'scanAsync' is retried until the interface is available. I could sit in a loop waiting for the interface to become available first but it makes no difference - also this only happens 1 in 10 times so it would slow things down unnecessarily 9 times out of  10.

start
```diff
@@ -2,12 +2,10 @@
 
 export DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket
 
-if [ ! -f /data/network.config ]; then
-	node src/app.js
-else
-	cp /data/network.config /host/var/lib/connman/network.config
-fi
+sleep 1 # Delay needed to avoid DBUS introspection errors
```
josephroberts: The problem is that without that delay it falls over [here](https://github.com/resin-io/resin-wifi-connect/blob/network-manager-support/src/systemd.coffee#L4)  so I am not sure how I can check for/catch the error because it occurs during the creation of a member variable - I guess I could create an init function for systemd? 

I think the `export....` line should be in bash because its only needed whilst resin-wifi-connect runs in the user container which is gonna change with the next version.

josephroberts: to be honest I am not sure why the error occurs, I think the only way to catch it inside the app and remove the need for a delay would be to wrap [this](https://github.com/resin-io/resin-wifi-connect/blob/network-manager-support/src/systemd.coffee#L4) in an init function that retries on error.

src/systemd.coffee
Page-: Ok, it was unclear that it can only have one element, that's a weird interface

PR #61, opened 2016-12-12
Removed duplicate access points from ssid list

### LINKED ISSUE #59: SSID list shows multiple entries for the same AP.

josephroberts:
Fixes #59

### DIFF
--- a/src/wifi-scan.coffee
+++ b/src/wifi-scan.coffee
@@ -1,5 +1,6 @@
 Promise = require 'bluebird'
 child_process = Promise.promisifyAll require 'child_process'
+_ = require 'lodash'
 
 config = require './config'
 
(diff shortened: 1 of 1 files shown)


src/wifi-scan.coffee
```diff
@@ -32,8 +32,25 @@ exports.scanAsync = ->
...
 			networks.push(details)
 
-		# sort by signal strength
-		return networks.sort((a, b) -> b.signal - a.signal)
+		# sort by signal strength and remove duplicates
+		networks = networks.sort((a, b) -> a.signal - b.signal)
+		networks = removeDuplicates(networks, 'ssid')
+		networks = networks.sort((a, b) -> b.signal - a.signal)
+
+		return networks
```
CameronDiver: This can be done with 
```
ns = _.sortedUniqBy(networks, 'ssid');
```

Page-: That function is a `_.uniqBy` that's optimised for presorted arrays, and we don't sort by ssid so it's not quite right, but I do love the idea of replacing the custom code with lodash!
My take on this is:
```coffee
_(networks).orderBy('signal', 'desc').uniqBy('ssid').sortBy('signal').value()
```

PR #65, opened 2017-01-11
Edison support

josephroberts:
Added support for the `intel-edison`
Added code to allow easy addition of device types in future

Note to self - ~~final test~~, release next version and ~~update example project~~


src/hotspot.coffee
```diff
@@ -5,18 +5,21 @@ execAsync = Promise.promisify(exec)
...
 	if started
 		return Promise.resolve()
 
 	started = true
 
-	console.log('Stopping service, starting hotspot')
+	console.log('Starting hotspot')
 
-	manager.stop()
+	modprobe.ap(device)
```
CameronDiver: This could be a more descriptive name.

CameronDiver: This function could be named more descriptively.

Dockerfile.template
```diff
@@ -1,5 +1,7 @@
 FROM resin/%%RESIN_MACHINE_NAME%%-node:6
 
+ENV DEVICE_TYPE=%%RESIN_MACHINE_NAME%%
```
josephroberts: Could I set this to `RESIN_DEVICE_TYPE`, would it still be white listed on older supervisors that don't have this set by default? @pcarranzav

src/modprobe.coffee
```diff
@@ -0,0 +1,25 @@
+Promise = require 'bluebird'
+{ spawn, exec } = require 'child_process'
+execAsync = Promise.promisify(exec)
+
+exports.ap = (device) ->
+	switch device
+		when 'intel-edison', 'edison'
+			run('modprobe -r bcm4334x', 'modprobe bcm4334x op_mode=2 firmware_path="firmware/intel-edison/fw_bcmdhd.bin" nvram_path="firmware/intel-edison/bcmdhd.cal"')
```
CameronDiver: Rather than a one case switch it might be better here to do

```
if device is 'intel-edison' or device is 'edison'
    run(...)
```

src/modprobe.coffee
```diff
@@ -0,0 +1,25 @@
...
+run = (disable, enable) ->
+	console.log('Loading kernel module: ' + enable)
+	Promise.delay(1000)
+	.then ->
+		execAsync(disable)
+	.catch(->)
+	.delay(1000)
+	.then ->
+		execAsync(enable)
+	.catch(->)
```
CameronDiver: Only one catch needed per promise chain here, might be worth displaying an error if they ever get called as well. At the moment if the modprobe fails it will do so silently.

src/modprobe.coffee
```diff
@@ -0,0 +1,25 @@
...
+			run('modprobe -r bcm4334x', 'modprobe bcm4334x op_mode=2 firmware_path="firmware/intel-edison/fw_bcmdhd.bin" nvram_path="firmware/intel-edison/bcmdhd.cal"')
+
+exports.normal = (device) ->
+	switch device
+		when 'intel-edison', 'edison'
+			run('modprobe -r bcm4334x', 'modprobe bcm4334x firmware_path="firmware/intel-edison/fw_bcmdhd.bin" nvram_path="firmware/intel-edison/bcmdhd.cal"')
+
+run = (disable, enable) ->
+	console.log('Loading kernel module: ' + enable)
+	Promise.delay(1000)
```
CameronDiver: I don't think you'll need a delay here, and possibly won't further down the chain. I think you could alleviate the delays by creating a function that resolves the promise on exec completion:

```
execPromise = (command) ->
    new Promise (resolve, reject) ->
         child = exec(command)
         child.addListener('error', reject)
         child.addListener('exit', resolve)
```
and then use it like:
```
execPromise(disable)
.then -> execPromise(enable)
.then -> console.log('done')
```

src/modprobe.coffee
josephroberts: there will be multiple devices eventually - all the artiks' for example.

src/modprobe.coffee
josephroberts: We have 2 catches to ensure each modprobe command runs

src/modprobe.coffee
```diff
@@ -0,0 +1,25 @@
...
+exports.normal = (device) ->
+	switch device
+		when 'intel-edison', 'edison'
+			run('modprobe -r bcm4334x', 'modprobe bcm4334x firmware_path="firmware/intel-edison/fw_bcmdhd.bin" nvram_path="firmware/intel-edison/bcmdhd.cal"')
+
+run = (disable, enable) ->
+	console.log('Loading kernel module: ' + enable)
+	execPromise(disable)
+	.then ->
+		execPromise(enable)
```
CameronDiver: It might be worth adding the catches back in, based on your reply to the comment about them - to ensure both modprobes are called. That is if you still want to be calling the second modprobe if the first one fails.

src/modprobe.coffee
josephroberts: It does not work, the delays are needed :(

src/modprobe.coffee
josephroberts: actually the catches are not needed.

Issue #71, opened 2017-03-02
Wifi does not come back up when the device is rebooted from within the application.


josephroberts:
Reported on the forums [here](https://forums.resin.io/t/rpi3-wifi-not-connecting-after-a-reboot-from-the-api/615)

Resin OS 1.24.1
Supervisor 2.8.3

josephroberts:
I managed to reproduce this after several tries.

There is a 15 second timeout during which connman has to connect to the configured network, this is great when a network has not already been configured because it means the configuration loop is fairly quick in the event a user enters the wrong password, makes a typo etc. However, if the network has already been configured `resin-wifi-connect` will exit once the timeout has elapsed and in some cases the timeout is too short. Throwing this error:
```
Clear disabled
Using connman.service
Credentials found
Connecting
Error: Timed out
    at Timeout._onTimeout (/usr/src/app/src/connman.js:78:25)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)
Not retrying
Exiting
```

josephroberts:
I tried increasing the timeout which didn't help, so I changed `resin-wifi-connect` to wait until connman is up before attempting to connect. This seems to work

PR #78, opened 2017-03-07
Increased timeout to 30 seconds

josephroberts:
Fiixes #71


src/app.coffee
```diff
@@ -54,7 +54,7 @@ run = ->
 			hotspot.stop(manager, device)
 			.then ->
 				console.log('Connecting')
-				manager.connect(15000)
+				manager.connect(30000)
```
CameronDiver: I would consider exporting this magic number to a const variable. In addition if it optionally pulled in from an env var, it would mean it could be changed without changing the code, for example;
```coffeescript
# 30 s is the default, override with RWC_CONNECT_TIMEOUT env var
CONNECT_TIMEOUT = process.env.RWC_CONNECT_TIMEOUT ? 30000
....
manager.connect(CONNECT_TIMEOUT)
```

PR #79, opened 2017-03-07
Removed magic timeout number and wait for service to start up before …

### LINKED ISSUE #71: Wifi does not come back up when the device is rebooted from within the application.

josephroberts:
…attempting to connect

Fixes #71

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -6,6 +6,7 @@ hotspot = require './hotspot'
 networkManager = require './networkManager'
 systemd = require './systemd'
 wifiScan = require './wifi-scan'
+config = require './config'
 
 app = express()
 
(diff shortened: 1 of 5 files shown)


src/app.coffee
```diff
@@ -117,6 +118,8 @@ systemd.exists('NetworkManager.service')
 	.then (setup) ->
 		if setup
 			retry = false
+.then ->
+	manager.ready()
```
CameronDiver: This can be 
```
.then(manager.ready)
```

josephroberts: Nope, it can't

PR #94, opened 2017-05-04
Server

josephroberts:


src/main.rs
```diff
@@ -36,81 +73,137 @@ fn main() {
...
+    let index = wifi_state.access_points.iter().position(|ref ap| ap.ssid == ssid).unwrap();
+    match connection::create(&wifi_state.manager,
+                             &wifi_state.device,
+                             &wifi_state.access_points[index],
+                             &password,
+                             10) {
+        Ok(connection) => {
+            // TODO: how do we respond and exit?
+            Ok(Response::with(status::Ok))
+            process::exit(0);
```
majorz: Here an implementation that we can use for exiting the program from a request: http://stackoverflow.com/a/31589035

majorz: Here an implementation that we can use for exiting the program from a request: http://stackoverflow.com/a/31589035

src/main.rs
```diff
@@ -36,81 +73,137 @@ fn main() {
...
+        manager: manager,
+        device: device,
+        access_points: access_points,
+        hotspot_connection: hotspot_connection,
+    };
+
+    // TODO: is this the best way to implement a timeout?
+    thread::spawn(move || {
+        start_server(wifi_state);
+    });
```
majorz: Maybe we need to pass the timeout value as a parameter in the WiFiState and use it in the `connect` handler with `connection::create`?

josephroberts: this timeout is meant to stop resin-wifi-connect if no wifi credentials are added for a period of time. It is different to the connection::create, enable etc. timeouts

majorz: In that case we may integrate a timer thread (like this one: https://github.com/Yoric/timer.rs) together with the Condvar mechanism for exiting the process from the second comment.

cmr:
Overall, looks good!

src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct CliOptions {
+    pub interface: Option<String>,
+    pub ssid: String,
+    pub password: Option<String>,
+    pub timeout: u64,
+}
+
+pub fn parse_cli_options() -> CliOptions {
+    let matches = App::new("resin-wifi-connect")
+        .version("0.1.0")
```
majorz: Nice!

rustfmt.toml
```diff
@@ -0,0 +1,7 @@
+use_try_shorthand = true
```
majorz: Right, I did not apply those, since we have not finalized them. I will add them once we merge the other PR.

majorz:
Closing in favor of a new PR

Issue #99, opened 2017-06-11
 Error: No introspectable


pcriadoperez:
Hi,

I am getting this error when flashing a Pi Zero W:
Clear is disabled
11.06.17 18:44:02 (-0400) Device type is raspberry-pi
11.06.17 18:44:02 (-0400) { Error: No introspectable
11.06.17 18:44:02 (-0400)     at /usr/src/app/resin-wifi-connect/node_modules/dbus/lib/bus.js:101:15
11.06.17 18:44:02 (-0400)   cause:
11.06.17 18:44:02 (-0400)    Error: No introspectable
11.06.17 18:44:02 (-0400)        at /usr/src/app/resin-wifi-connect/node_modules/dbus/lib/bus.js:101:15,
11.06.17 18:44:02 (-0400)   isOperational: true }
11.06.17 18:44:02 (-0400) Exiting

If I start from the terminal it works ( $node src/app.js --clear=true )

pcriadoperez:
It works if I increase the sleep time to from 1 to 30
And  set --clear=true

majorz:
We added support for `START_SLEEP` environment variable to the start script in v2.0.8 as a short term solution. The error does not exist in the ongoing rewrite of the application in the Rust language.

PR #101, opened 2017-06-13
Configurable sleep in the start script

majorz:
Change-Type: patch
Connects-To: #99
---- Autogenerated Waffleboard Connection: Connects to #99


majorz:
> Can we add some instructions to the readme please

I added a FAQ entry for the new environment variable. Please, let me know if this needs some improvement.

Issue #104, opened 2017-06-28
Rust language reimplementation


majorz:
Reimplement the existing functionality using the Rust programming language.

PR #105, opened 2017-06-28
Rust language reimplementation

majorz:
Connects-To: #104
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #104


majorz:
This new PR that includes all of the Rust implementation code in a single commit.

This is the new flow of the application:

New Flow

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+  - cargo install-update -i rustfmt
+  - if [ $TRAVIS_RUST_VERSION == nightly ]; then
+      cargo install-update -i clippy;
+    fi
+script:
+  - cargo fmt -- --write-mode=diff
+  - if [ $TRAVIS_RUST_VERSION == nightly ]; then
+      cargo clippy -- -D warnings;
+    fi
+  - cargo build
```
josephroberts: You need to update the readme. We also need to release the 2.x version with the warnings we spoke about in the arch call before this can be merged

Cargo.lock
```diff
@@ -0,0 +1,1003 @@
+[root]
+name = "resin-wifi-connect"
```
josephroberts: should a binary have this lock file? I thought it was only recommended for libraries

src/network.rs
```diff
@@ -0,0 +1,380 @@
...
+                        return exit_with_error(
+                            exit_tx,
+                            dnsmasq,
+                            format!("Stopping the hotspot failed: {}", e),
+                        );
+                    }
+                    hotspot_connection = None;
+                }
+
+                let access_points = match get_access_points(&device) {
```
josephroberts: there is some code duplication, for example this section is repeated on line 95

src/server.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+impl typemap::Key for RequestSharedState {
+    type Value = RequestSharedState;
+}
+
+unsafe impl Send for RequestSharedState {}
+unsafe impl Sync for RequestSharedState {}
+
+#[derive(Debug)]
```
josephroberts: can we remove this, you have the display trait implemented below

src/network.rs
josephroberts: I have used [this](https://github.com/mibk/dupl) for go code - not sure if there is something similar for rust. I am surprised clippy does not catch this to be honest

src/dnsmasq.rs
```diff
@@ -0,0 +1,23 @@
+use std::process::{Child, Command};
+
+use network_manager::Device;
+
+use config::Config;
+
+pub fn start_dnsmasq(config: &Config, device: &Device) -> Result<Child, String> {
```
abrodersen: `String` errors make me sad. Have you looked at using `error_chain`?

src/network.rs
```diff
@@ -0,0 +1,380 @@
...
+                            e.description()
+                        ),
+                    );
+                }
+            },
+            NetworkCommand::Connect {
+                ssid,
+                password,
+            } => {
+                if hotspot_connection.is_some() {
```
abrodersen: I think this should be an `if let Some(value)`

src/network.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    let exit_tx_server = exit_tx.clone();
+    let gateway = config.gateway;
+    thread::spawn(move || { start_server(gateway, server_rx, network_tx, exit_tx_server); });
+
+    'main_loop: loop {
+        let command = match network_rx.recv() {
+            Ok(command) => command,
+            Err(e) => {
+                return exit_with_error(
```
abrodersen: I think instead of calling `exit_with_error` all over the place, it might make sense to convert all this main loop logic into a dispatch function that returns a `Result`. You can then use the handy `?` operator inside the function to avoid all the `match`ing. The main loop can then check for an `Err` result and send the exit command.

src/network.rs
```diff
@@ -0,0 +1,380 @@
...
+        .map(|ap| ap.ssid().as_str().unwrap().to_string())
+        .collect()
+}
+
+fn create_hotspot(
+    device: &Device,
+    ssid: &str,
+    gateway: &Ipv4Addr,
+    password: &Option<&str>,
+) -> Result<Connection, String> {
```
abrodersen: Is this the `rustfmt` approved way of formatting a function? It looks really wonky to me.

src/server.rs
```diff
@@ -0,0 +1,229 @@
...
+struct RedirectMiddleware;
+
+impl AfterMiddleware for RedirectMiddleware {
+    fn catch(&self, req: &mut Request, err: IronError) -> IronResult<Response> {
+        let gateway = {
+            let request_state = get_request_state!(req);
+            format!("{}", request_state.gateway)
+        };
+
+        let host = req.headers.get::<headers::Host>().unwrap();
```
abrodersen: Could a client crash the server by not sending a `Host` header?

src/server.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        if host.hostname != gateway {
+            let url = Url::parse(&format!("http://{}/", gateway)).unwrap();
```
abrodersen: Same with sending a malformed `Url`

Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "resin-wifi-connect"
+version = "0.1.0"
+authors = ["Joe Roberts <joe@resin.io>"]
```
majorz: :D

Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "resin-wifi-connect"
+version = "0.1.0"
```
majorz: will place a matching "2.0.8"

Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+mount = "0.3"
+router = "0.5"
+serde_json = "1.0"
+persistent = "0.3"
+params = "0.6"
+log = "0.3"
+env_logger = "0.4"
+
+[dependencies.network_manager]
+git = "https://github.com/resin-io-modules/network_manager"
```
majorz: Sounds good - let's leave that for a next PR

src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+                .long("clear")
+                .value_name("CLEAR")
+                .help("Clear saved Wi-Fi credentials")
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let interface: Option<String> = matches.value_of("interface").map_or_else(
+        || {
+            env::var("PORTAL_INTERFACE").ok()
```
majorz: If no device is specified, we default to the first Wi-Fi device. Do we really need to default to 'wlan0'? For example on my laptop the internal one is called 'wlo1' and this one makes more sense as a default one for my system. Or I am missing something?

src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    let ssid: String = matches.value_of("ssid").map_or_else(
+        || {
+            env::var("PORTAL_SSID").unwrap_or_else(|_| DEFAULT_SSID.to_string())
+        },
+        String::from,
+    );
+
+    let password: Option<String> = matches.value_of("password").map_or_else(
+        || {
+            env::var("PORTAL_PASSPHRASE").ok()
```
majorz: It is the same as in the current implementation - the default is for creating a open hotspot.

.travis.yml
majorz: As we talked on the chat about those items - the PR will go into `104-rust`, not `master` and we will make the changes in consecutive PRs.

Cargo.lock
majorz: It is the other way around: http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

Cargo.toml
josephroberts: @majorz this will be `3.0.0` - the last coffeescript version we release will be `2.0.8`

src/config.rs
josephroberts: no, that makes total sense. ignore me :)

src/config.rs
josephroberts: ok

src/dnsmasq.rs
majorz: We discussed that earlier on during one of the network_manager PRs, but left it for a later stage. @josephroberts, what's your opinion on the improved error reporting - should we pursue that before release or after?

josephroberts: I think we should release now and have this as one of the first things we "fix"

src/network.rs
majorz: Yeah, I am not happy with that long function either. It definitely needs a better design and I was thinking that we can address that in the next development phase. Joe had some ideas about incorporating some form of state machine as well. @josephroberts, what's your opinion on this - should we try to come up with a better design for the network related thread function, or we should address that at a post-3.0 phase?

src/network.rs
majorz: nice catch!

src/network.rs
josephroberts: address it later on

src/network.rs
majorz: Yes, this is how `rustfmt` is configured now for us to avoid right-side drifting (I am not really sure what is the current `rustfmt` default, since they improved many aspects of those recently). We need to catch up with `rustfmt` soon, since they introduced a new crate that runs on nightly only, which will be the maintained version: http://www.ncameron.org/blog/rustfmt-changes/

src/server.rs
majorz: Another nice catch, will fix that.

src/server.rs
majorz: Will fix that too :)

src/dnsmasq.rs
majorz: Definitely it will take more time to get it done in both repos. We will address it later on then.

Cargo.toml
majorz: @josephroberts, it will be bumped to `3.0.0` by VersionBot automatically when `104-rust` lands in `master`. So I think it is better to keep the Cargo version with the changelog one in sync.

src/network.rs
majorz: There are actually some small changes (like a missing `let` in one of the cases) and maybe this is why clippy does not catch it. It is still a duplication, but I suggest that we leave it like that for now and address it all at once when we refactor that whole function with a better design.

src/server.rs
majorz: I tried to remove it and the compiler complains with `required by "std::error::Error"`.

src/server.rs
majorz: I looked at this one and it will actually never fail, since the gateway is already parsed Ipv4Addr instance.

majorz:
I updated the source code and recommitted with all the changes discussed above.

Additionally I added support for:
1. Starting the network manager service if stopped for some reason
2. Checking for Internet connectivity after connection to access point is established - when new connection is created using the hotspot and webserver. If connectivity is not established the connection will be deleted and the webserver will continue to run waiting for new credentials to be added.

src/config.rs
```diff
@@ -0,0 +1,125 @@
+use clap::{Arg, App};
+
+use std::env;
+use std::net::Ipv4Addr;
+use std::str::FromStr;
+
+const DEFAULT_GATEWAY: &str = "192.168.42.1";
+const DEFAULT_DHCP_RANGE: &str = "192.168.42.2,192.168.42.254";
+const DEFAULT_SSID: &str = "ResinAP";
+const DEFAULT_TIMEOUT: &str = "15000";
```
josephroberts: We should specify that this timeout is in ms, either with a comment or by naming the variable `DEFAULT_TIMEOUT_MS`

src/config.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub password: Option<String>,
+    pub clear: bool,
+    pub gateway: Ipv4Addr,
+    pub dhcp_range: String,
+    pub timeout: u64,
+}
+
+pub fn get_config() -> Config {
+    let matches = App::new(env!("CARGO_PKG_NAME"))
+        .version(env!("CARGO_PKG_VERSION"))
```
josephroberts: I didn't know about `CARGO_PKG_...` - very nice :)

src/server.rs
josephroberts: hmm, any idea why?

majorz: The Error trait is defined as dependent on the Debug and Display traits: `pub trait Error: Debug + Display {...}`.

src/config.rs
majorz: Good point, I changed it to `DEFAULT_TIMEOUT_MS`.

majorz:
In addition to the `DEFAULT_TIMEOUT_MS` change I added one fix for the travis builds:

When clippy was compiled with an older Rust nightly (like from the previous day), but no new clippy version was found, then clippy is not working with the new Rust nightly (because it links to the compiled Rust binary libs). In that case although a new clippy version is not available, we still need to recompile the current clippy version towards the new nightly. The check I did is run `cargo clippy -V` and then if the exit code is non-zero something is wrong (like the described case) and we force reinstall.

PR #114, opened 2017-09-13
Rust language reimplementation

majorz:
Change-Type: major
Connects-To: #104
---- Autogenerated Waffleboard Connection: Connects to #104


.gitignore
```diff
@@ -35,3 +35,10 @@ src/public/bower_components
 
 # VSCode
 .vscode
+
```
josephroberts: Do you think we should clean this out so that it contains rust stuff only?
https://github.com/github/gitignore/blob/master/Rust.gitignore

Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "resin-wifi-connect"
+version = "2.0.9"
```
josephroberts: version needs bumping (or does VB do it now?)

Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+mount = "0.3"
+router = "0.5"
+serde_json = "1.0"
+persistent = "0.3"
+params = "0.6"
+log = "0.3"
+env_logger = "0.4"
+
+[dependencies.network_manager]
+git = "https://github.com/resin-io-modules/network_manager"
```
josephroberts: we should really publish network_manager as a crate and remove this.

Dockerfile.template
```diff
@@ -1,34 +1,20 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-node:6
+FROM resin/%%RESIN_MACHINE_NAME%%-debian
```
josephroberts: can we use a `slim` image?

Dockerfile.template
```diff
@@ -1,34 +1,20 @@
...
-		dnsmasq \
-		hostapd \
-		iproute2 \
-		iw \
-		libdbus-1-dev \
-		libexpat-dev \
-		rfkill \
-	&& rm -rf /var/lib/apt/lists/*
+    && apt-get install -y dnsmasq \
+    && rm -rf /var/lib/apt/lists/*
```
josephroberts: `apt-get clean && rm -rf /var/lib/apt/lists/*`

Dockerfile.template
```diff
@@ -1,34 +1,20 @@
...
 
 RUN mkdir -p /usr/src/app/
```
josephroberts: this line is not needed

Dockerfile.template
```diff
@@ -1,34 +1,20 @@
...
-
-COPY bower.json .bowerrc /usr/src/app/
-
-RUN ./node_modules/.bin/bower --allow-root install \
-	&& ./node_modules/.bin/bower --allow-root cache clean
-
-COPY . /usr/src/app/
-
-RUN ./node_modules/.bin/coffee -c ./src
+COPY start /usr/src/app/
```
josephroberts: you can do `COPY start .`

Dockerfile.template
```diff
@@ -1,34 +1,20 @@
...
 
 CMD bash start
```
josephroberts: Should be `CMD ["bash","start"]`

README.md
```diff
@@ -7,18 +7,18 @@ This is a [resin.io](http://resin.io) application. Check out our [Getting Starte
 
 This project is meant to be integrated as part of a larger application (that is, _your_ application). An example on how to use this on a Python project can be found [here](https://github.com/resin-io-projects/resin-wifi-connect-example).
```
josephroberts: this project needs to be updated

README.md
```diff
@@ -7,18 +7,18 @@ This is a [resin.io](http://resin.io) application. Check out our [Getting Starte
...
+This is a Rust application, but your app can be any language/framework you want as long as you install it properly - if you need help, check out our [Dockerfile guide](https://docs.resin.io/deployment/dockerfile/). This project uses a Resin feature called "Dockerfile template", this means that the base image is chosen depending on the architecture, specified by the `%%RESIN_MACHINE_NAME%%` variable (see [line 1](./Dockerfile.template#L1) in the template).
 
 ## How it works
-This app interacts with either Connman or Network Manager in Resin's base OS. First it checks whether WiFi has been previously configured, if it has it attempts to connect to the configured network, if it hasn't it opens an Access Point to which you can connect using a laptop or mobile phone and input new WiFi credentials.
+This app interacts with Network Manager in Resin's base OS. First it checks whether WiFi has been previously configured, if it has it attempts to connect to the configured network, if it hasn't it opens an Access Point to which you can connect using a laptop or mobile phone and input new WiFi credentials.
```
josephroberts: change `base OS` to `host OS`

src/config.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub ui_path: PathBuf,
+}
+
+pub fn get_config() -> Config {
+    let matches = App::new(env!("CARGO_PKG_NAME"))
+        .version(env!("CARGO_PKG_VERSION"))
+        .author(env!("CARGO_PKG_AUTHORS"))
+        .about(env!("CARGO_PKG_DESCRIPTION"))
+        .arg(
+            Arg::with_name("interface")
```
josephroberts: It would be good to explain what these args do and how to set them in the readme. Can they all be set with args and with environment variables? Which takes precedence?

src/config.rs
```diff
@@ -0,0 +1,144 @@
...
+            env::var("PORTAL_PASSPHRASE").ok()
+        },
+        |v| Some(v.to_string()),
+    );
+
+    let clear = matches.value_of("clear").map_or(true, |v| !(v == "false"));
+
+    let gateway = Ipv4Addr::from_str(&matches.value_of("gateway").map_or_else(
+        || {
+
```
josephroberts: weird newline here, did rustfmt do that?

.gitignore
majorz: I was wondering that as well and it looks like a good idea. I will clean it up.

Cargo.toml
majorz: Yes, VB will handle the version bumping. However now I remember that I need to set this up in the versionist.conf.js file, so I will fix this.

Cargo.toml
majorz: :D

Dockerfile.template
majorz: I will check it.

src/config.rs
majorz: Yes, it is confusing indeed and I will document all of them in the README. We can use both args and environment variables, but the arguments have higher precedence.

josephroberts: I think it might be worth stripping the readme back to the basics needed to make this work and putting all the other stuff into the wiki. what do you think?

majorz: Right, I will put that on the Wiki instead. And we may link to it from the README.

Dockerfile.template
majorz: No `slim` versions are provided for `debian` since it is already a slim version.

src/config.rs
majorz: Yes. After we release have to upgrade to latest rustfmt, as they changed it to run on nightly only and all new development is happening now in "rustfmt-nightly".

src/server.rs
```diff
@@ -0,0 +1,230 @@
...
+    network_tx: Sender<NetworkCommand>,
+    exit_tx: Sender<ExitResult>,
+}
+
+impl typemap::Key for RequestSharedState {
+    type Value = RequestSharedState;
+}
+
+unsafe impl Send for RequestSharedState {}
+unsafe impl Sync for RequestSharedState {}
```
abrodersen: Why are these traits specified? Should they be automatically implemented if all the members of `RequestSharedState` implement it?

majorz: That is a proper assertion. I changed the implementation to use a Mutex for the shared state and now we do not have to declare `Sync` anymore. The `Send` trait is automatically implemented and should not have been defined.

src/config.rs
majorz: Wiki page started here: https://github.com/resin-io/resin-wifi-connect/wiki/Command-Line-Arguments. Added a link in the README.

README.md
majorz: PR for the project created and approved by Joe.

Issue #116, opened 2017-10-06
Raspbian Stretch install script


majorz:
Implement a bash script for easy installation on Raspbian Stretch. This requires downloading the upcoming v3 from Github releases, disabling dhcpcd and installing and activating NetworkManager.

shawnipoo:
Running either install script fails. Is there a fresh install script? Thank you.

lurch:
I guess this can be closed now?

majorz:
Implemented in #117.

PR #117, opened 2017-10-06
Raspbian Stretch install script

majorz:
Implements a bash script for easy installation on Raspbian Stretch. If necessary it disables dhcpcd and installs NetworkManager to act as network manager instead. Then it downloads the upcoming v3 from Github releases and extracts it in the current working dir.

Connects-To: #116
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #116


majorz:
@abrodersen, I think the reason Alex wanted a script was to get it out quicker - soon after the v3 release. I remember packaging was mentioned at some point, but this is probably something that we won't look into very soon.

Issue #118, opened 2017-10-06
README note for the coming-soon v3


majorz:
v3 was not released when we deprecated v2 (#112), because we had to remake our toolchains to support ARMv6. To avoid confusion we need a README note that v3 is coming soon.

PR #119, opened 2017-10-06
README note for the coming-soon v3

majorz:
Connects-To: #118
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #118


majorz:
This should have been a `patch` change type, which I realized when versionbot merged with the new version tag. Should not cause any issues though.

Issue #145, opened 2017-11-07
Support for multiple connection profiles


andrewn:
If I'm reading the v3 docs correctly, `WiFi Connect` will only store one network to connect to? The previous behaviour was to maintain a list of networks and credentials. If it could not connect to them, it would start AP mode.

For a user of my project, I think it might be quite annoying to have to re-enter credentials when moving between wifi networks, is there any way/plans to store a list of networks again?

majorz:
We have not changed the behavior in v3 compared to v2. When `--clear=false` is set, the list with connections will be preserved, but the application will not enter AP/captive portal mode. With `--clear=true` it will remove all defined connections and enter AP mode. I just double checked that by running v2 with both options and I confirm that this was the behavior before. On the positive side is that I verified that the new version is much more stable and correct ;)

I agree that the current behavior is a bit limited and confusing and I personally look forward to expanding the functionality and adding support for preserving already entered credentials and entering AP mode at the same time.

majorz:
I am will follow up with more information this week.

andrewn:
Thanks @majorz. I was using v1 before and skipped to v3 so I didn't know about this behaviour.

For my use case I would definitely like the behaviour of being able to enter AP mode whilst preserving previous credentials.

Is this something that's planned? Do you think it's something that would be straightforward to add? I've never used Rust but I'm willing to give it a go and submit a PR.

majorz:
@andrewn I will put it to the top of the priority list. It is the most required feature anyway. What is most important is to define how we will expand the command line usage. Either with new definitions or by expanding the —clear argument values. In a way that more use cases are handled nicely and easy. The —clear argument is not a fine cli api and I am more on the side for deprecating it, but leaving it there till v4. I will be glad to hear your opinion and ideas.

andrewn:
Thanks for the quick reply!

There are a few use cases:

a. Automatically start the AP whenever no Wi-Fi network is joined
b. Force AP mode when `wifi-connect` starts without clearing networks
c. Clear networks

I agree about deprecating `--clear`. I think that the 3 use cases above would be separate CLI arguments and would give `wifi-connect` users more control for different situations.

andrewn:
@majorz Just [looking at the code](https://github.com/resin-io/resin-wifi-connect/blob/master/src/network.rs#L234-L241). It seems that if you have previously never connected to a network, then an AP is created, which is a bit like use case (a) above.

majorz:
[...]

PR #157, opened 2018-01-17
Support for multiple connection profiles

majorz:
This PR is a major version update of WiFi Connect targeting support for multiple connection profiles. The previous version supported defining only a single connection profile through WiFi Connect, which limited moving the device between locations, etc.

To support multiple connection profiles the application flow is now simplified and excludes unnecessary functionality:
1. Trying to activate existing connections: this is handled by NetworkManager automatically.
2. Deleting all existing wireless connections: this is out of WiFi Connect functionality scope and can be achieved in other ways if necessary.
3. Not exiting the application if Internet connectivity is not established after connecting to the WiFi network. Now the application exits even if no Internet connectivity is established after connecting to the wireless network, as we should not make assumptions about the connectivity state of the network a user wants to connect to. Displaying the connectivity state on exit is kept though.

Since this is a major release not compatible with v3 there was room for other improvements as well:
1. Introducing `--activity-timeout` argument for exiting the application in case a user has not connected to the captive portal for the specified amount of time. The default is no activity timeout. This is needed in use cases where WiFi Connect can block the start shell script when Internet connectivity is temporary unavailable, e.g. Boombeastic does a google ping to determine whether to execute WiFi Connect and without this argument the main application will not start in case the device is restarted and there is temporary connectivity issue.
2. The old `--timeout` argument which was about specifying amount of seconds to be waited for connection to be established is now removed. This was introduced in v3, but there was not a single case where this was really needed. In the future if such cases arise, we can introduce something similar again and implement it better.
3. The directory containing the web UI assets of the application is renamed from `public` to `ui`.
4. The example start script is renamed from `start` to `scripts/start.sh` and now it includes different example conditions for executing WiFi Connect before the application. The default condition is "Is there an active WiFi connection?" by checking the exit code of `iwgetid -r`. For this reason `wireless-tools` is installed in the Dockerfile.template. 
5. Installation instructions section in the Readme including installing on Raspbian/Debian Stretch. 
6. WiFi Connect logo and a new How It Works image.
7. The code in network.rs is refactored for reducing the cyclomatic complexity.

Change-Type: major
Connects-To: #145
---- Autogenerated Waffleboard Connection: Connects to #145


drewcovi:
@majorz thanks for this! what are the "other ways" now to clear connections?

majorz:
@drewcovi it is safe to leave unused connection profiles. You can also clean them up manually from the host OS with `nmcli c` and `nmcli c delete <UUID>`. If you have many of those and would like to clear them regularly you may automate `nmcli` or use the NetworkManager D-Bus API directly. If you are using Python I recommend the python-networkmanager package.

Issue #176, opened 2018-02-08
Better README section for running on top of resinOS


majorz:
The section for running on top of resinOS is too short and has to be revisited and expanded a bit.

mbardelmeijer:
Yes please :) The `resin-wifi-connect-example` repository seems outdated for v4.0, and installing via the `raspbian-install.sh` trough the Dockerfile doesn't seem to work for me.

majorz:
@mbardelmeijer thanks for pointing that. The documentation is confusing for newcomers indeed. The `raspbian-install.sh` script should not be used when running on top of resinOS. You can use the included `Dockerfile.template` in the repository directly.

This line here downloads and extracts latest release of the application: https://github.com/resin-io/resin-wifi-connect/blob/master/Dockerfile.template#L12.

Please let me know if you run into any problems when trying to run the application.

mbardelmeijer:
Hey, thanks for the extra info!

I'm running into a problem with that unfortuantaly. We are running the `resin/raspberrypi3-debian:stretch` base image, which does not include the `network-manager` by default. After installing and enabling it the device will go offline and have to reinstall.

Any tips? :)

majorz:
@mbardelmeijer Right, the base image does not include NetworkManager (this is the one that is run as a Docker container). And it should not run it, since it will generally collide with the host OS (resinOS). NetworkManager runs in the host OS though, and WiFi Connect communicates with it using D-Bus.

In the example start.sh script that you may find in the scripts directory you will see the following line:
`export DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket`

This basically helps WiFi Connect to communicate with NetworkManager that runs on the host OS.

From the Dockerfile.template the only important dependency that has to be installed in the container is `dnsmasq`, since WiFi Connect needs it to setup the captive portal: https://github.com/resin-io/resin-wifi-connect/blob/master/Dockerfile.template

BTW you may clone the resin-wifi-connect repo itself and push it directly as a resin.io application and it will work out of the box with `resin/raspberrypi3-debian` (it should be Stretch marked as latest image version as well).

Please let me know if you have further questions.

mbardelmeijer:
Awesome, after setting the `DBUS_SYSTEM_BUS_ADDRESS` everything worked perfectly! Thanks a lot for your help 😄

majorz:
[...]

PR #178, opened 2018-02-08
Start NetworkManager after it's installed

mbardelmeijer:
We ran into the issue with the 4.0 version that NetworkManager was installed with the `raspbian-install.sh` script, but not started & enabled. This made the installation fail. 

Hope this can be merged or some other (perhaps prettier) solution :)


majorz:
@mbardelmeijer thanks a lot for working on this. I will review it in a few hours and get back to you.

majorz:
We solved this previously in a separate issue - #176, so I will close the PR now, as we discovered that it is not a problem with the install script itself.

Issue #122, opened 2017-10-06
Prompt to elevate privileges or warn about non-privileged execution


majorz:
The application needs elevated privileges to be able to create the captive portal hotspot. We can prompt to elevate privileges or warn the user about non-privileged execution and exit.

PR #193, opened 2018-03-14
Require root privileges

majorz:
Elevating the privileges from inside the application is very complicated, so we are replacing the current process with a `sudo` version of the same command using `execvp` internally.

This works especially well on Raspbian, where the default `pi` user has passwordless `sudo`. When `sudo` is not passwordless, the user is asked for password with the regular `sudo` prompt.

Change-Type: patch
Connects-To: #122
---- Autogenerated Waffleboard Connection: Connects to #122


majorz:
We discussed this outside of GH, so I will repost.

First is to clarify why we need to run WiFi Connect as root. NetworkManager has different policies across systems - in a desktop environment it will allow creating the hotspot with a regular user, but on Raspbian for example it does not. We have similar restrictions about opening port 80 and running dnsmasq. Since it will be quite a lot of work to elevate privileges from inside the app, e.g. using policykit, then we are left with the option of requiring the app to be run as root only.

I explored what I could find on GitHub as solutions and I think the sudo one is closest to what my original idea was. The other option would be to just tell the user that he need to run the application as root and exit. But since he will do sudo wifi-connect anyway, we can just do it for him directly. Works quite nicely on Raspbian, since the default pi user there is in the sudoers list and has passwordless sudo enabled.

Although not very common there are tools that use this approach. My assumption is that people are not aware that this can be done quite seamlessly, so they just print an error and exit. I guess the closest to this is how sudo can be invoked in a shell script - the shell script will show the sudo prompt and continue further. This one is different as it does not invoke only commands, but uses services that need privileges.

Possible drawbacks need to be investigated before merging though. Otherwise we will just error out with a root requirement message and exit.

shaunmulligan:
@majorz I think I agree with @abrodersen on this one, I would be more comfortable just printing an error and telling the user to start wifi-connect with sudo. In this way they feel they have more control, even if it is the same level of control :)

majorz:
Thanks guys, we are printing an error now.

PR #202, opened 2018-04-04
Remove blank/duplicate SSIDs from dropdown list

handsomgenious:
Blank entries will appear for hidden SSIDs, and duplicate entries will appear for multiple APs broadcasting the same SSID.

Small AND conditional wrapping the append call, preventing either/both of these scenarios.


majorz:
@handsomgenious thanks for the PR! All of that definitely makes sense. I will move this type of filtering in the Rust part of the application, so that the JSON response is filtered when it comes from the web server. I will do it next week.

PR #211, opened 2018-04-30
Fix for nmcli -t g error

jgentes:
Error is "Option '--terse' requires specifying '--fields'"

It's also worth noting that using `nmcli` requires the installation of `network-manager`, however I doubt it's worth adding the dependency unless people want to use this command.


majorz:
Thanks for the PR, I will check it out as soon as I find time. The reason for the error is that different nmcli versions have different CLI arguments. I need to check the exact details.

Issue #220, opened 2018-06-08
Error: Getting the NetworkManager service state failed


tonyho:
I download the binary/ELF and run the following version files at my ARM linux, but it is not working.
Used version: wifi-connect-v4.1.1-linux-armv7hf.tar.gz

The relative log is: 
```
# ./wifi-connect 
Error: Getting the NetworkManager service state failed
  caused by: Service
# NetworkManager --version
1.10.2
# ps ax | grep Network
  260 root     /usr/sbin/NetworkManager
  528 root     grep Network
# 
```

My board is:
arm v7l Linux 4.1.15.  And the networkmanager is listening the port:
`
unix  2      [ ACC ]     STREAM     LISTENING       3344 /var/run/NetworkManager/private-dhcp

`

Also I can use the nmcli to control the Networkmanger.

majorz:
Are you running wifi-connect in a resin OS container or another distro? Asking because this error usually occurs when running in a resin OS container, where `DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket` should be exported before running the application.

majorz:
The root cause is that wifi-connect does not find the NetworkManager service through the systemd D-Bus interface. Is it a systemd enabled system?

majorz:
If so, probably we need to put a warning there and continue operation without exiting with an error.

tonyho:
Thanks majorz. Our system is running with SysVInit init script instead of systemd, also, not the resin OS container, we build the rootfs using the Yocto.

About the D-Bus, I guess it's exposed for the nmcli and nmtui can interact with the NetworkManager. 

So, can this wifi-connect work on such system?  Or what should I do to make it work on the system?

majorz:
Alright, so this is what I expected. Currently wifi-connect expects systemd to be active. I will fix this in the next release, so that you may use it under SysVInit. I should be able to get it done today.

majorz:
This is now fixed in v4.1.2.

PR #221, opened 2018-06-11
Do not fail on systems without systemd

majorz:
WiFi Connect gets the NetworkManager service state through the systemd D-Bus API. If systemd is not active on the system, the application exits with error. Now it just prints an info message that the NetworkManager service cannot be retrieved.

Connects-to: #220
Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #220


src/network.rs
```diff
@@ -426,8 +426,12 @@ fn wait_for_connectivity(manager: &NetworkManager, timeout: u64) -> Result<bool>
 }
 
 pub fn start_network_manager_service() -> Result<()> {
-    let state =
-        NetworkManager::get_service_state().chain_err(|| ErrorKind::NetworkManagerServiceState)?;
+    let state = if let Ok(state) = NetworkManager::get_service_state() {
```
dfunckt: I think you can use `match` over `NetworkManager::get_service_state()` result? I think it is more idiomatic, isn't it?

majorz: Yes, it is probably a bit more idiomatic. I recommitted with a `match` statement.

PR #249, opened 2018-10-30
Pending changes from O.S. Systems for merging

otavio:
We made a small image for a demo and made use of this nice tool. We ended making few changes here which seems to be good to be send upstream, so here it goes...


otavio:
It has been a while since we sent this PR and we did not get feedback. Is it possible one of maintainers to review it and ask for the needed changes?

PR #251, opened 2018-11-02
bugfix RUN step fails because curl does not redirect

mpalmier74:
Build of docker image fails on step 5/7:

`Step 5/7 : RUN curl https://api.github.com/repos/resin-io/resin-wifi-connect/releases/latest -s     | grep -hoP 'browser_download_url": "\K.*rpi\.tar\.gz'     | xargs -n1 curl -Ls     | tar -xvz -C /usr/src/app
`
This fails because resource referenced by curl has been moved to a new URL and curl does not redirect to new URL. Option "-L" to curl enables redirect and allow the build to continue and succeed

NOTE: I learned about -L from this [link](https://stackoverflow.com/questions/18474690/is-there-a-way-to-follow-redirects-with-command-line-curl)

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -9,7 +9,7 @@ RUN apt-get update \
 
 WORKDIR /usr/src/app
 
-RUN curl https://api.github.com/repos/resin-io/resin-wifi-connect/releases/latest -s \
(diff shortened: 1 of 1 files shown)


zrzka:
Thanks! It's no longer necessary, because the instructions were updated (`resin-io` -> `balena-io`) and there's no redirect involved. Closing.

Issue #265, opened 2019-02-15
Blank & duplicate SSID entries


t16n:
In the network list I can see multiple entries for the same SSIDs. Also I see empty entries.

https://github.com/balena-io/wifi-connect/pull/61
and 
https://github.com/balena-io/wifi-connect/pull/202

tried to fix it, but i think the issue still remains.

majorz:
@t16n right, I believe the code has changed from the time those issues were created. Probably the easiest way for fixing is filtering duplicated ssid members in the JavaScript here: https://github.com/balena-io/wifi-connect/blob/6547a4a77c76d0ec903033f44a177f312c2831fe/ui/js/index.js#L21

I will handle those cases when I get back to transitioning the project to the new NetworkManager Rust library I have been working on, but I am not sure when I will be able to dedicate the time for this because of the higher priorities with some other projects we have.

astekardis:
Has there been any progress on this? I'm trying to use wifi-connect and am seeing duplicate networks appear in the SSID list (none of the duplicate networks are the one that I need to connect to).
<img width="910" alt="Screen Shot 2019-11-05 at 11 10 00 AM" src="https://user-images.githubusercontent.com/56287775/68237005-279bcd00-ffbb-11e9-84ea-26c57d57c090.png">

HughWarrington:
This affects me too. Any update?

majorz:
Both of those issues are now fixed in v4.4.3.

PR #269, opened 2019-02-20
Remove duplicate and empty SSID entries.

t16n:
Change-Type: patch
Connects-To: #265

### DIFF
--- a/ui/js/index.js
+++ b/ui/js/index.js
@@ -19,7 +19,7 @@ $(function(){
 		} else {
 			networks = JSON.parse(data);
 			$.each(networks, function(i, val){
-				if((val !== '') && ($('#ssid-select option[val="' + val + '"]').length === 0)){
+				if((val.ssid !== '') && ($('#ssid-select option[val="' + val.ssid + '"]').length === 0)){
 					$('#ssid-select').append(
 						$('<option>')
 							.text(val.ssid)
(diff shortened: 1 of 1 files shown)


ui/js/index.js
```diff
@@ -19,12 +19,14 @@ $(function(){
 		} else {
 			networks = JSON.parse(data);
 			$.each(networks, function(i, val){
-				$('#ssid-select').append(
-					$('<option>')
-						.text(val.ssid)
-						.attr('val', val.ssid)
-						.attr('data-security', val.security)
-				);
+				if((val !== '') && ($('#ssid-select option[val="' + val + '"]').length === 0)){
```
maruchronos: This change doesn't solve the problem, you must add `val.ssid` instead of `val` in order to make it work.
```if((val.ssid !== '') && ($('#ssid-select option[val="' + val.ssid + '"]').length === 0)){```

zrzka:
Hi, thanks for the PR! I think that we should handle this elsewhere - I mean, not the presentation layer, but in the [networks](https://github.com/balena-io/wifi-connect/blob/master/src/server.rs#L173-L195) endpoint. Presentation layer should display everything what's returned from the `networks` endpoint. What do you think? If you're not familiar with Rust, I can do it.

majorz:
Closing in favor of #339.

PR #275, opened 2019-03-28
Change the base image to balenalib

horia-delicoti:


Dockerfile.template
```diff
@@ -1,11 +1,8 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-debian
+FROM balenalib/%%RESIN_MACHINE_NAME%%-debian
 
 ENV INITSYSTEM on
```
imrehg: @horia-delicoti this line does not do anything on `balenalib` base images. Likely need to remove it as well.

imrehg:
@majorz I see that the rust test fails, due to formatting check fail. While it is not required, would that be something to fix up in some subsequent PR? or the rust tests are not in use in general yet?

Dockerfile.template
horia-delicoti: @imrehg Done.

majorz:
@imrehg those Rust tests are some new addition to ResinCI which probably we need to remove as we already do those kinds of checks on the CircleCI front. Speaking of CircleCI it does not show on the status anymore...

PR #276, opened 2019-03-28
Add CODEOWNERS

imrehg:
So that the listed code-owers will always be added to PR reviews, etc.

Change-type: patch
Signed-off-by: Gergely Imreh <imrehg@gmail.com>

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*	@majorz @zrzka


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*	@majorz @zrzka
```
hedss: Need newline :)

imrehg: Does it?

imrehg: @hedss that's very silly...
(coupled with `nano`'s 4.0 changes that no automatic newline added :P )

PR #280, opened 2019-04-02
Ci token checks

### LINKED ISSUE #278: Broken CI deployment of releases
Yesterday's release was pushed, but the CI failed to deploy the binaries on GitHub. Working on it now.

majorz:
Adding checks for the GitHub token used by the CI, so that deployment does not break in he future.

Also one additional commit for simplifying the CI config: 
do not manually define GitHub user and repo since those are already available as CircleCI env vars.

Closes #278

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,30 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: Get the username associated with the GitHub token
+          command: |
+            USER_URL=https://api.github.com/user
+            GITHUB_TOKEN_USER=$(curl -s -H "Authorization: token $GITHUB_TOKEN" "$USER_URL" | jq -r .login)
+            echo "GITHUB_TOKEN_USER=$GITHUB_TOKEN_USER"
+            echo "export GITHUB_TOKEN_USER=$GITHUB_TOKEN_USER" >> $BASH_ENV
+
+      - run:
+          name: Test whether the username was retrieved successfully
+          command: test -n "$GITHUB_TOKEN_USER" && test "$GITHUB_TOKEN_USER" != null
+
+      - run:
+          name: Get the repository permissions for the GitHub token
+          command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -73,17 +73,37 @@ jobs:
...
+          name: Get the username associated with the GitHub token
+          command: |
+            USER_URL=https://api.github.com/user
+            GITHUB_TOKEN_USER=$(curl -s -H "Authorization: token $GITHUB_TOKEN" "$USER_URL" | jq -r .login)
+            echo "GITHUB_TOKEN_USER=$GITHUB_TOKEN_USER"
+            echo "export GITHUB_TOKEN_USER=$GITHUB_TOKEN_USER" >> $BASH_ENV
+
+      - run:
+          name: Test whether the username was retrieved successfully
+          command: test -n "$GITHUB_TOKEN_USER" && test "$GITHUB_TOKEN_USER" != null
```
zrzka: @majorz is this step necessary? I mean, `jq -e` exits if the `.login` key doesn't exist or if the value is `false` or `null`. Or even better, here's an example of what `jq` can do:

```
jq -r '.login | if type != "string" or . == "" then halt_error(1) else . end
```

This one exits with 1 if the value doesn't exist, if it's not a string or if it's empty.

zrzka: Taking back this comment. The error won't be propagated from `$()`.

majorz: I structured it this way for simplicity and readability. Do you object strongly to leaving it as it is? I do not see much value in changing it.

majorz: I particular I do not like the jq usage above - looks like more error prone this way than not.

zrzka: No objection, I took my comment back. Because error propagating from `$()` will be way more noisy.

majorz:
@zrzka One idea that I have is to move all the CI specifics in a Rust CLI application with subcommands that will have integration tests and cleaner code. Currently all the CI code including the deployment script is too complex to be implemented as shell script.

The `jq` improvement above indicates this as well. It is too of a complex task.

It all works alright as currently as it is, but if we need to expand it, we need to definitely consider it.

PR #287, opened 2019-05-10
Replace UI with React and Rendition

sradevski:
To anyone reviewing, since the rust images we use to build wifi-connect don't have node and npm, we decided to just include the `build` folder for this PR so the build process improvements are separated from the migration to React.

See: https://www.flowdock.com/app/rulemotion/r-product/threads/A79Rsm9ich5RBUfKu22kEAzksVb
Change-type: minor
Signed-off-by: Stevche Radevski <stevche@balena.io>


LucianBuzzo:
@sradevski What is the difference in bundle size after this change? I was under the impression that we want3ed to keep wifi-connect as lightweight as possible.

sradevski:
@LucianBuzzo the build folder is 4.8MB (without compression). If the source maps are removed (which is now on my todo list) it drops to 1.1MB. Originally it was around 500kb, so it isn't a big difference.

LucianBuzzo:
@majorz Is ~ 600kb an acceptable increase in bundle size for the UI?

majorz:
@sradevski @LucianBuzzo 600KB more will not be a problem.

ui/src/components/Notifications.tsx
```diff
@@ -0,0 +1,40 @@
...
+  error
+}: {
+  hasAvailableNetworks: boolean;
+  isConnecting: boolean;
+  error: string;
+}) => {
+  return (
+    <>
+      {isConnecting && (
+        <Alert info>
```
LucianBuzzo: The notifications need a little padding/margin around them, at the moment they touch each other, and the edge of the screen:
Screen Shot 2019-06-17 at 09 25 51

ui/src/components/Notifications.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <Txt.span>
+            Your device will soon be online. If connection is unsuccessful, the
+            Access Point will be back up in a few minutes, and reloading this
+            page will allow you to try again.
+          </Txt.span>
+        </Alert>
+      )}
+      {!hasAvailableNetworks && (
+        <Alert warning>
+          <Txt.span>No wifi networks available. </Txt.span>
```
LucianBuzzo: There is a trailing space here:
```suggestion
          <Txt.span>No wifi networks available.</Txt.span>
```

ui/src/components/NetworkInfoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+        </Heading.h3>
+
+        <Box width={['80%', '60%', '40%']} my={3}>
+          <Txt>SSID</Txt>
+          <DropDownButton
+            width={'100%'}
+            joined
+            label={<Txt>{this.state.formData.ssid || 'Select SSID'}</Txt>}
+          >
+            {this.props.availableNetworks.map(x => (
```
LucianBuzzo: Can you use a more semantic variable than `x`?
```suggestion
            {this.props.availableNetworks.map(availableNetwork => (
```

ui/src/components/NetworkInfoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <Box width={['80%', '60%', '40%']} my={3}>
+          <Txt>Passphrase</Txt>
+          <CustomInput
+            type="password"
+            width={'100%'}
+            onChange={e => this.change('passphrase', e.target.value)}
+          />
+        </Box>
+
+        <Button primary onClick={this.submit}>
```
LucianBuzzo: Can you wrap these inputs in a `form` so that hitting `Enter` submits the form?

ui/src/components/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+    isFetchingNetworks: true,
+    availableNetworks: [],
+    error: ''
+  };
+
+  componentDidMount() {
+    fetch('/networks')
+      .then(data => {
+        if (data.status !== 200) {
+          throw new Error();
```
LucianBuzzo: Can you check for an error message here?

ui/src/components/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        return data.json();
+      })
+      .then((data: Network[]) => {
+        this.setState({ availableNetworks: data });
+      })
+      .catch((e: Error) => {
+        console.log(e);
+        this.setState({ error: 'Failed to fetch available networks' });
```
LucianBuzzo: Is it worth adding the error message to this?
```suggestion
        this.setState({ error: `Failed to fetch available networks: ${error.message || error}` });
```

ui/src/components/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  render() {
+    return (
+      <Provider>
+        <Navbar
+          brand={
+            <img
+              src={logo}
+              style={{ height: 30 }}
+              className="App-logo"
```
LucianBuzzo: Is this class necessary?

sradevski: It's not, a remnant from CRA

ui/src/components/Notifications.tsx
sradevski: That's intentional and shown in the UI, but I replaced it with `&nbsp;` for clarity.

sradevski:
> @sradevski @LucianBuzzo 600KB more will not be a problem.

After updating to the latest rendition with tree-shaking included, it dropped to 400KB

ui/src/components/App.tsx
sradevski: I thought just logging it in the console would be enough, but sure, it is easier for users to see what the problem is.

ui/src/components/NetworkInfoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+	},
+});
+
+const isEnterpriseNetwork = (
+	networks: Network[],
+	selectedNetworkSsid?: string,
+) => {
+	const selectedNetwork = networks.filter(
+		(x) => selectedNetworkSsid === x.ssid,
+	)[0];
```
JSReds, line 60: here you could also use a `find` instead of the `filter`, so you don't have to take the first one

iamsolankiamit:
any reason to include build in commit?

sradevski:
@iamsolankiamit the base images don't include node and there are few of them that need to be included, so we decided to just include the build now, and then separately work on the build pipeline.

ui/src/components/NetworkInfoForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+			emphasized: true,
+		},
+	},
+});
+
+const isEnterpriseNetwork = (
+	networks: Network[],
+	selectedNetworkSsid?: string,
+) => {
+	const selectedNetwork = networks.filter(
```
iamsolankiamit, line 58: we can use networks.some and add both condition there

ui/src/img/logo.svg
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
```
iamsolankiamit, line 1: we can compress these with https://jakearchibald.github.io/svgomg/

ui/src/index.css
```diff
@@ -0,0 +1,12 @@
+body {
```
iamsolankiamit, line 1: can be added in styled components GlobalStyles.

PR #313, opened 2019-09-18
Delete only the right AP profile on start

### LINKED ISSUE #312: Do not delete all AP profiles on start
Currently WiFi Connect deletes all AP connection profiles on start. It should delete only the one it creates.

majorz:
Fixes #312

Change-type: patch

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -211,7 +211,7 @@ impl NetworkCommandHandler {
     }
 
     fn connect(&mut self, ssid: &str, identity: &str, passphrase: &str) -> Result<bool> {
-        delete_connection_if_exists(&self.manager, ssid);
+        delete_existing_connections_to_same_network(&self.manager, ssid);
 
         if let Some(ref connection) = self.portal_connection {
             stop_portal(connection, &self.config)?;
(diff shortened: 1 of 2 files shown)


src/network.rs
```diff
@@ -508,18 +508,22 @@ pub fn start_network_manager_service() -> Result<()> {
...
     for connection in connections {
-        if &connection.settings().kind == "802-11-wireless" && &connection.settings().mode == "ap" {
-            debug!(
-                "Deleting access point connection profile: {:?}",
-                connection.settings().ssid,
-            );
-            connection.delete()?;
+        if let Ok(connection_ssid) = connection.settings().ssid.as_str() {
+            if &connection.settings().kind == "802-11-wireless"
+                && &connection.settings().mode == "ap" && connection_ssid == ssid
```
imrehg: Would it be good to clarify here, what this this check doing? Something like "we check for profies for access points that were likely created by us, and not to touch other profiles" or similar (didn't think too much about it, but an explanation seems to be good)

majorz: > How would this work if someone changes the SSID in the config, would that result in leftover connection profiles? Just wondering.

A left-over connection profile will be something rarely seen - a power cycle or sigkill, where the process did not had a chance to exit cleanly.

I could not think of a better fix, but if you have better ideas?

majorz: Hm, I do not know why that comment went over there...

majorz: As for adding a comment there - I prefer not to leave comments, but refactor the code if something is not clear enough, e.g. change function names or do function extraction.

majorz: So maybe changing `delete_access_point_profile` to `delete_existing_access_point_profile`?

imrehg: @majorz is there any problem commenting on the "intent"? I don't like code that has no comments at all (especially Rust where comments are first class citizens)

Yeah, the rename might help, but `delete_existing_access_point_profile` also not straightforward (of course you can only delete "existing" profile, and the nuance is what we delete and what we don't delete.

I am not trying to change your style (okay, a little bit:),  and maybe this is rather a `nit:`, but I feel **quite uneasy** sending review through, where the code is not necessarily clear clear (like this multilayer conditionals). Here you would have to read and think what the code does, instead of having any help from the author to understand the intent. Anyways.

majorz: @imrehg refactored the code to cover the code concerns. The conditionals are living now in separate functions. Function names and log entries improved for describing the intent clearly.

PR #316, opened 2019-10-23
Support Buster and future Raspbian versions in the installer

### LINKED ISSUE #314: Error trying to install on last Raspbian version
Error trying to install on last Raspbian version:

`WiFi Connect Raspbian Installer: Distribution not based on Debian 9 (stretch)`

image

majorz:
Reverse installer logic to exclude Jessie only from allowing installation.

Fixes #314
Change-type: patch

### DIFF
--- a/scripts/raspbian-install.sh
+++ b/scripts/raspbian-install.sh
@@ -69,8 +69,8 @@ check_os_version() {
         _version=$(grep -oP 'VERSION="\K[^"]+' /etc/os-release)
     fi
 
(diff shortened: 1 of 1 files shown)


majorz:
> can we add a check for Jessie and then fail gracefully saying its not supported.

Yes, that will be even better. I thought only about forward compatibility. Will reverse the if statement and check whether it is a Jessie version.

majorz:
@shaunmulligan done

PR #318, opened 2019-11-05
Remove duplicates when calling `get_access_points_ssids`

Page-:
Change-type: patch

Linked to #265 

This solves an annoyance of having duplicate ssids returned when there are multiple networks available with the same ssid, eg
```
# nmcli device wifi list
IN-USE  SSID             MODE   CHAN  RATE        SIGNAL  BARS  SECURITY 
        BTWifi-with-FON  Infra  1     195 Mbit/s  100     ****  --       
        BTWifi-with-FON  Infra  11    195 Mbit/s  32      **    --   
```
where `BTWifi-with-FON` is a public network that is broadcast from multiple routers

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -387,10 +387,15 @@ fn get_access_points_impl(device: &Device) -> Result<Vec<AccessPoint>> {
 }
 
 fn get_access_points_ssids(access_points: &[AccessPoint]) -> Vec<&str> {
-    access_points
+    let mut ssids: Vec<&str> = access_points
         .iter()
         .map(|ap| ap.ssid().as_str().unwrap())
-        .collect()
+        .collect();
+    // Sort and then dedup to remove all duplicate ssids, which come from
+    // the same ssid on multiple channels/frequencies
+    ssids.sort();
+    ssids.dedup();
+    ssids
 }
 
 fn get_networks(access_points: &[AccessPoint]) -> Vec<Network> {


majorz:
@Page- we need to preserve the order of the SSIDs, so that they are sorted by signal strength from top to bottom. This is the easiest way I could think of - https://play.rust-lang.org/?gist=c901f433bb3e45035cca56638e7e4c09. If you find something shorter, even better.

And I think we may modify the comment to "Remove all duplicate SSIDs by preserving signal strength order". The "which come from the same ssid on multiple channels/frequencies" part is not very accurate and we can omit it altogether I think.

Thanks again for the PR!

HughWarrington:
Is there anything I can do to help this along? Unfortunately I don't seem to be able to log in to the CI system above to see why the tests are failing.

majorz:
I will try to find time in the next days to update the PR with the needed changes and release a fix for this.

HughWarrington:
That's great, thank you.

Although in my opinion, the PR might be fine as it stands -- in reply to your remark earlier 

> we need to preserve the order of the SSIDs, so that they are sorted by signal strength from top to bottom.

I think it is a good thing to sort the SSIDs alphabetically given there's no indication of signal strength in the dropdown. To the user, without being prompted with the signal strength the sort order may just look random. On top of this if you already know the SSID you are looking for it is hard to find in the list when it's not sorted alphabetically. I've been using this code in a large shared office where there are 20-30 access points (and each on several frequencies) and the network I want to connect to is far from the strongest.

majorz:
Closing in favor of #339.

Issue #327, opened 2020-01-14
Scan only shows 1 ssid some of the time


baileysage:
Running on a raspi3 B v1.2 on Buster. Not using wpa. 

Sometimes scan will only return one ssid, even if all of the available networks show up in iwlist just before launching. 

We ARE setting a network priority on the ssid that always shows up, but otherwise running according to the README.

marclennox:
I get this also.

majorz:
That's interesting. What happens if you run `nmcli d wifi rescan` just before running `wifi-connect`? `iwlist` uses the older wireless extensions for Linux where NetworkManager is talking to the kernel through the newer netlink mechanism (the `iw` utility uses that as well). `iw dev wlan0 scan` is also an interesting output.

marclennox:
I will test this out and let you know.

baileysage:
We get the same single network from `nmcli d wifi rescan` that we see in wifi-connect.

Interestingly, running `sudo iw wlan0 scan` will kick something into gear and then nmcli and wifi-connect will see all available networks.  BUT, then, connecting to the network will fail and restart the AP. Then, wifi-connect will show the network we attempted to connect to, allow us to select it, enter credentials, and connect as normal. I have `sudo journalctl -fu NetworkManager` output I can sanitize and share if that would be helpful.

majorz:
Thanks for the feedback! The logs from NetworkManager would not reveal anything most probably, but I could be wrong.

There may be some kernel errors. Recently I have seen more and more instabilities with the RPi WiFi driver/firmware. Could you please also check the `dmesg` output and let me know if you see any crashes there.

NetworkManager delegates currently scanning to the wpa_supplicant. In the future probably that will be done by default by iwd - a new wireless daemon from Intel, which has a much cleaner codebase, better automated testing and is using more kernel routines. This one unfortunately is not compatible with the current RPi 4.19 kernel, so switching to it is not an option. The current combination of buggy drivers/firmware and wpa_supplicant issues makes things quite unstable. Other WiFi chips are even worse.

It would be best if I can reproduce this issue and I am going to do a round of testing. Are there any particular steps that you would recommend? You are probably running latest Raspbian Lite (Buster) - correct me if I am wrong. The error should also be affecting our balenaOS since we are running the same kernel and drivers (although patch versions could be vary).

Also if you are running Raspbian Lite, are you running it with full system upgrade or a stock one from September?

baileysage:
[...]

Issue #225, opened 2018-06-25
Refresh available WiFi Network list from captive portal


sako0938:
For whatever reason the user's WiFi network might not be up when wifi-connect is initially run. In addition there are situations where wifi-connect only see's a single wifi network when scanning. To assist with my own debugging, and for user experience, I feel this button should be something available on the captive portal. It would be nice to easily determine if it just did a bad scan, which happens often, even to my own PC vs the WiFi network actually not being there. Would prevent a lot of debugging tips like "reboot and see if it works this time"

majorz:
I recommend using the `--activity-timeout` command line argument for cases where the network is unavailable when the application is started. WiFi Connect will exit after the specified number of seconds if no user activity is detected. This will unblock the rest of the start script - e.g. starting the main application.

The issue with the visible single wifi network is a known one that I plan to address very soon.

The refresh button is a very nice idea though. I think we should add this anyway. Thanks for the feedback!

sako0938:
Thanks. I'm already using the activity-timeout, but not having network available isn't my real problem it just comes up in my testing a lot. Great to hear on the other fronts however, I'll keep checking back!

charliesneath:
@majorz Will `wifi-connect` ever attempt to rescan if there are no networks available?

silvia-eio:
Any idea of when we will have the refresh button?

silvia-eio:
Also, the --activity timeout seems not to be working for me... 
If I turn on the device in an area with networks "a", "b" and "c" available, it scans and shows them appropriately. 
If I don't connect to any of those networks, and move to a different area, with networks "x", "y" and "z" available, the Wifi-Connect access point will never refresh the network list - will keep showing only "a", "b" and "c" even though they are no longer within reach. 
Is there a way to fix that, or is there something wrong with the way I used activity timeout in the code below?

```
while true; do
   
    echo `date` " 3. Is there Internet connectivity via a google ping?"
    wget "http://clients3.google.com/generate_204?" -O /dev/null 2>&1 | grep "204 No Content" > /dev/null
    
    if [ $? -eq 0 ]; then
        printf 'Skipping WiFi Connect\n'
    else
        printf 'Starting WiFi Connect\n'
        ./wifi-connect --portal-ssid "Camera ${RESIN_DEVICE_NAME_AT_INIT}" --activity-timeout 60
    fi
    
    echo "Use control-c to quit this script"
    sleep 10
done
```

[...]

PR #354, opened 2020-10-01
Add a refresh networks button

sradevski:
@majorz Is this all that is needed? As far as I could tell `/network` refetches the networks, unless there is some caching in place that I missed.

Connects-to: #225
Change-type: minor
Signed-off-by: Stevche Radevski <stevche@balena.io>

### DIFF
--- a/ui/src/components/App.tsx
+++ b/ui/src/components/App.tsx
@@ -39,7 +39,8 @@ const App = () => {
 		[],
 	);
 
-	React.useEffect(() => {
+	const fetchNetworks = () => {
+		setIsFetchingNetworks(true);
 		fetch('/networks')
 			.then((data) => {
 				if (data.status !== 200) {
@@ -55,6 +56,10 @@ const App = () => {
 			.finally(() => {
 				setIsFetchingNetworks(false);
 			});
+	};
+
+	React.useEffect(() => {
+		fetchNetworks();
 	}, []);
 
 	const onConnect = (data: NetworkInfo) => {
@@ -92,6 +97,7 @@ const App = () => {
 					error={error}
 				/>
 				<NetworkInfoForm
+					fetchNetworks={fetchNetworks}
 					availableNetworks={availableNetworks}
 					onSubmit={onConnect}
 				/>
--- a/ui/src/components/NetworkInfoForm.tsx
+++ b/ui/src/components/NetworkInfoForm.tsx
@@ -1,7 +1,8 @@
 import { JSONSchema7 as JSONSchema } from 'json-schema';
 import * as React from 'react';
-import { Flex, Form, Heading, RenditionUiSchema } from 'rendition';
+import { Button, Flex, Form, Heading, RenditionUiSchema } from 'rendition';
 import { Network, NetworkInfo } from './App';
+import { RefreshIcon } from './RefreshIcon';
 
 const getSchema = (availableNetworks: Network[]): JSONSchema => ({
 	type: 'object',
@@ -61,11 +62,13 @@ const isEnterpriseNetwork = (
 };
 
 interface NetworkInfoFormProps {
+	fetchNetworks: () => void;
 	availableNetworks: Network[];
 	onSubmit: (data: NetworkInfo) => void;
 }
 
 export const NetworkInfoForm = ({
+	fetchNetworks,
 	availableNetworks,
 	onSubmit,
 }: NetworkInfoFormProps) => {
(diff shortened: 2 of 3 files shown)


maggie0002:
Fantastic, just what is needed!

From a UI perspective, perhaps makes more sense to have the refresh button clearly associated with the SSID box, alongside or above-right? And maybe 'refresh' should be something like 'rescan'?

sradevski:
@maggie0002 I agree with you on both points, but there are some (intentional) limitations on how we render forms, so adding the button next to the SSID box is not easily achievable without hacking around it. I've updated the text to say "Rescan" though, that's much better.

maggie0002:
Looks great, I think the 'rescan' text provides the clarity that was needed by itself. 👍

sradevski:
@gregmihalko Thanks for the thorough test and explanation. Honestly, I just went ahead and implemented the UI changes while I had some free time, and I was hoping @majorz can have a look at what needs to be done on the backend part as I have very little working knowledge of Rust. Once he has the time to check it out he can do the necessary changes, and we can get this merged. Once that's done, it's trivial to improve the UX, which is why I wanted to keep the changes here to the minimum.

gregmihalko:
@majorz let me know if there's anything else I might be able to test or try to implement further.

maggie0002:
This line won't be necessary if the refresh button is implemented: 

https://github.com/balena-io/wifi-connect/blob/56f61632f60fd1977112d7fa080fa0268f858e3e/ui/src/components/Notifications.tsx#L29

maggie0002:
> any updates when this PR could be merged?

I too am keeping a close eye on this one, would be great to see it implemented.

That said, when I think about it I realise that it is perhaps not this pull request that I am looking for, but a fix to the reason I need the pull request. On occasion when I load the device all the available networks in the area are not listed properly. Subsequently, I came to want a refresh networks button. Ideally though, the issue of the networks not all being listed would be resolved first and foremost. The circumstances where I would boot the device, then move closer to a network and want to refresh the list is very rare, and a hard reboot there not a major issue. The refresh network button then was only really to resolve a bug for me.

I'm going to have a look at implement this fix: https://github.com/balena-os/wifi-connect/issues/327#issuecomment-577794377

I also hope that that fix is implemented into Wi-Fi connect first and foremost, to reduce the need for a refresh button in the first place.

immiz:
> > any updates when this PR could be merged?
> 
> I too am keeping a close eye on this one, would be great to see it implemented.
> 
> That said, when I think about it I realise that it is perhaps not this pull request that I am looking for, but a fix to the reason I need the pull request. On occasion when I load the device all the available networks in the area are not listed properly. Subsequently, I came to want a refresh networks button. Ideally though, the issue of the networks not all being listed would be resolved first and foremost. The circumstances where I would boot the device, then move closer to a network and want to refresh the list is very rare, and a hard reboot there not a major issue. The refresh network button then was only really to resolve a bug for me.
> 
> I'm going to have a look at implement this fix: [#327 (comment)](https://github.com/balena-os/wifi-connect/issues/327#issuecomment-577794377)
> 
> I also hope that that fix is implemented into Wi-Fi connect first and foremost, to reduce the need for a refresh button in the first place.

this makes sense - thanks for the update @maggie0002

majorz:
The problem with the PR is that scanning is not allowed by NetworkManager while the WiFi device is in AP mode. I am not sure what the proper solution for this would be, but it definitely needs a much deeper investigation.

iamsolankiamit:
@majorz would disabling the AP mode, rescanning and starting AP again if needed, solve the issue?

maggie0002:
> @majorz would disabling the AP mode, rescanning and starting AP again if needed, solve the issue?

I think the issue would be that the user would be connected via the access point. So you would lose connection to the user interface and the wifi hotspot, would need to reconnect again and refresh the page. May be better than having to restart the whole device, but also not ideal.

iamsolankiamit:
@maggie0002 i think a refresh option with a warning is better than restarting the device. I presume that switching between WiFi mode would be faster than restarting the device. This also helps people who might get confused as to why their WiFi is not being shown, even after moving the device closer in range/ restating the router etc. not being aware that the WiFi-connections are only scanned once.

maggie0002:
> @maggie0002 i think a refresh option with a warning is better than restarting the device. I presume that switching between WiFi mode would be faster than restarting the device. This also helps people who might get confused as to why their WiFi is not being shown, even after moving the device closer in range/ restating the router etc. not being aware that the WiFi-connections are only scanned once.

I think so too. 

According to this thread (https://forums.balena.io/t/wifi-connect-community-maintainers/331507) @majorz is working on updating some of the dependencies of Wi-Fi connect which with any luck will open up some other options and maybe resolve some issues. I imagine it will be hard to know what they may be until the updates are done but we are eagerly awaiting to see.

maggie0002:
Quick update on some experiences with refreshing network points. I have been using the IW package (seem to be getting more success with that than nmcli). Seems like on some devices you can refresh the networks list while the access point is live (tested on Raspberry Pi 4), and on other devices you cannot (tested on Orange Pi Zero LTS). When the refresh fails because the AP is up on the Orange Pi it returns an non-zero exit code. 

I ended up implementing an IW call on the page load and if the exit code return non-zero then disable the refresh networks button and prompt the user to restart the device, and if it returns 0 then allow the refresh network button.

PR #356, opened 2020-10-07
Update Dockerfile.template

### LINKED ISSUE #355: curl request to GH api might get rejected due to api rate limiting
https://github.com/balena-io/wifi-connect/blob/258656bf9b531ff64438ba502166840e48125e33/Dockerfile.template#L7

when run on balenaCloud could lead to GH API rate limiting. 

```
[wifi-connect]          {"message":"API rate limit exceeded for <REDACTED>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

A workaround is target releases directly at the cost of having to explicitly state the target version

curcuz:
fixes #355

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -8,6 +8,6 @@ ARG VERSION="4.4.0"
 RUN curl -Ls "https://github.com/balena-io/wifi-connect/releases/download/v$VERSION/wifi-connect-v$VERSION-linux-$BALENA_ARCH.tar.gz" \
   | tar -xvz -C  /usr/src/app/
 
-COPY ./start.sh .
+COPY scripts/start.sh .
 
 CMD ["bash", "start.sh"]
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,13 +1,12 @@
-FROM balenalib/%%RESIN_MACHINE_NAME%%-debian
+FROM balenalib/%%BALENA_MACHINE_NAME%%-debian
+ARG VERSION="4.4.0"
 
 RUN install_packages dnsmasq wireless-tools
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


Dockerfile.template
```diff
@@ -1,13 +1,12 @@
-FROM balenalib/%%RESIN_MACHINE_NAME%%-debian
+FROM balenalib/%%BALENA_MACHINE_NAME%%-debian
+ARG VERSION="4.4.0"
```
dfunckt, line 2: This is better placed later so that changing version doesn't trigger an unnecessary rerun of previous steps.

Dockerfile.template
```diff
@@ -1,14 +1,13 @@
...
-RUN curl https://api.github.com/repos/balena-io/wifi-connect/releases/latest -s \
-    | grep -hoP 'browser_download_url": "\K.*%%RESIN_ARCH%%\.tar\.gz' \
-    | xargs -n1 curl -Ls \
-    | tar -xvz -C /usr/src/app/
+ARG VERSION="4.4.0"
+RUN curl -Ls "https://github.com/balena-io/wifi-connect/releases/download/v$VERSION/wifi-connect-v$VERSION-linux-$BALENA_ARCH.tar.gz" \
+  | tar -xvz -C  /usr/src/app/
 
-COPY scripts/start.sh .
+COPY ./start.sh .
```
dfunckt, line 11: This is a weird change -- are you sure it's needed?

curcuz: it's a mistake, nice catch - I don't know how I ended up editing that line :S

dfunckt (approved):

PR #409, opened 2021-07-26
Include captive portal port on redirect

ab77:
* allows the use of non-default listening ports
* allows captive portal redirect when using non-standard listening port

Change-type: patch

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -19,6 +19,7 @@ use params::{FromValue, Params};
 use errors::*;
 use network::{NetworkCommand, NetworkCommandResponse};
 use exit::{exit, ExitResult};
+use config::Config;
 
 struct RequestSharedState {
     gateway: Ipv4Addr,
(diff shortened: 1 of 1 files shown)


ab77:
@majorz would also be awesome if we could `rm -rf .circle/` in the next few PRs..

src/server.rs
```diff
@@ -106,15 +107,16 @@ where
...
         let gateway = {
             let request_state = get_request_state!(req);
             format!("{}", request_state.gateway)
         };
+        let listening_port = config.listening_port;
 
         if let Some(host) = req.headers.get::<headers::Host>() {
             if host.hostname != gateway {
-                let url = Url::parse(&format!("http://{}/", gateway)).unwrap();
+                let url = Url::parse(&format!("http://{}:{}/", gateway, listening_port)).unwrap();
```
ab77, line 119: @majorz can you please help me out here with getting the listening port in here?

PaulAnnekov:
any news? great PR

maggie0002:
What does this do? Wifi connect already supports alternative ports as far as I know. And captive portal requests are a request made from the your phone, laptop or other device to a url set by the device manufacturer on port 80. Unless wifi connect was to listen on port 80 for captive portal requests I don’t see how it can achieve captive portal on an alternative port. And if it does use port 80, then that port is blocked for other apps so may as well just run the whole app on port 80.

majorz:
The way this currently works is that when a non-default listening port is specified another application will listen on port 80 and all port 80 requests will be routed to it, so this redirect handler will not be reached at all. The application that listens on port 80 has to implement a redirect like this.

maggie0002:
Here is a list I compiled of redirects for NGINX: https://github.com/LearnersBlock/learners-block/blob/main/frontend/config/nginx/conf/captive-portal.conf

My app listens on port 80, and wifi connect 8080. So I have to capture requests to these urls requested on port 80 by these different devices and then forward that request to my captive portal page. 

An alternative way to do it is to capture and forward all 404 page not found errors to your captive portal, which is more comprehensive that having to specify a list, but means you can’t have custom 404s and users may accidentally stumble along the captive portal page by mistake when entering a wrong url.

Issue #412, opened 2021-08-31
Captive portal tries to load a font from google fonts


PaulAnnekov:
UI uses [rendition](https://github.com/balena-os/wifi-connect/blob/9d8b773e3bb4e6448ea39e46e4d7c0b7f7f63f77/ui/package.json#L13) library which [tries to load ](https://github.com/balena-io-modules/rendition/blob/2285c864a8311ae10bc55661c2a1b32532d823fd/src/components/Provider/index.tsx#L43) a google font. 
As there is no internet, you see a loader for several seconds until browser understands resource is unavailable end cancels load. It's a bad UX. 

<img width="1025" src="https://user-images.githubusercontent.com/1203892/131578445-4d79513e-83fe-4a92-8091-411b24bbba7f.png">

maggie0002:
That would explain some behaviour, nice find. Would be a matter of storing the font file in the repo and point the interface to the local copy. 

Think @sradevski built the interface.

PR #418, opened 2022-01-06
Fix offline page attempts to load from google fonts

joeframbach:
Since the device has no network access, the font request will fail and delay the page load.

Connects-to: #412
Change-type: minor


joeframbach:
I'd like to know if there is an easier way to type the `PartialDeep<Theme>` expected by `Provider` since that type isn't exported by rendition.

joeframbach:
It looks like all PRs are failing this test, I'm not able to access the details. Likely unrelated to this PR in particular.

PR #476, opened 2023-02-17
Use Flowzone as a CI and upgrade network-manager crate

majorz:
This fixes a failure at start due to missing AP security flags and device
type enumerations in the network-manager crate.

Also update to using Rust 1.67.1 and Flowzone as a CI.


majorz:
@balena-ci I self-certify!

PR #478, opened 2023-02-20
Remove unused package-lock.json

majorz:
Change-type: patch


majorz:
@balena-ci I self-certify!

PR #492, opened 2023-06-09
Allow wifi-connect to refresh networks while connected

tomasgareau:
This introduces a quick-and-dirty way that we could allow `wifi-connect` to refresh its list of nearby APs without restarting the device or interrupting the captive portal. Not sure if this would work for all of `wifi-connect`'s use-cases but it has worked for us so I thought I'd put up a PR to see if there was any interest in this approach (or something like it).

We fire off a `wifiscanner::scan` on every call to `activate`. `wifiscanner` uses `iw` under the hood on Linux which means we can scan for nearby APs _while_ still providing the captive portal via NetworkManager.

We _could_ convert the resultant `wifiscanner::Wifi` to a `Network` but unfortunately `wifiscanner` doesn't currently parse security info from the output of `iw` :grimacing: Fortunately, this scan still seems to allow NetworkManager to discover new APs, so we can just refresh our list of access points and `get_networks` as usual.

This means that every call to `/networks` will kick off a new scan. This should allow #354 to "just work" since it would get an updated list of networks every time. Otherwise I've found I can also just refresh the captive portal page.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
+dependencies = [
+ "memchr 2.5.0",
+]
+
+[[package]]
 name = "ansi_term"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


meelash:
> We _could_ convert the resultant `wifiscanner::Wifi` to a `Network` but unfortunately `wifiscanner` doesn't currently parse security info from the output of `iw` 😬 Fortunately, this scan still seems to allow NetworkManager to discover new APs, so we can just refresh our list of access points and `get_networks` as usual.

This didn't work for me on a pi4 for whatever reason. NetworkManager was only returning the broadcasted AP, even after the iw scan. I therefore switched to using a pull request on wifiscanner that has security info and adding the parsing of the results to `Network`s.

I've created a new pull request of that here: https://github.com/balena-os/wifi-connect/pull/521

PR #515, opened 2023-11-07
Migrate to use Vite instead of Create React App

### LINKED ISSUE #514: Proposal: Replace create-react-app with Vite for Improved UI Development and DX
I am writing to propose a change in our development process that I believe will significantly improve our UI development speed and developer experience (DX). Specifically, I suggest that we replace create-react-app with Vite.

Vite offers several advantages over create-react-app:

Faster Development: Vite offers a faster development server start-up and updates the modules instantly when you make changes.
Improved DX: Vite has features like out-of-the-box support for TypeScript, JSX, CSS pre-processors, and more.
Optimized Builds: Vite uses Rollup for its production builds, which results in smaller and faster-to-load bundles.
Migrating from create-react-app to Vite is a straightforward process. Here is a [guide] that outlines the steps involved.

I am willing to work on this migration and believe that this change will greatly enhance our development process.

Thank you for considering this proposal.

thienandangthanh:
Fix #514


thienandangthanh:
I found out there's a problem that the ui only show blank page when running with wifi-connect binary.
I'll take a look to solve it.

PR #555, opened 2024-04-16
Enable building and publishing of binaries via flowzone

rahul-thakoor:
Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -3,18 +3,29 @@ name: Flowzone
 on:
   pull_request:
     types: [opened, synchronize, closed]
-    branches:
-      - 'main'
-      - 'master'
+    branches: [main, master]
+  # allow external contributions to use secrets within trusted code
+  pull_request_target:
+    types: [opened, synchronize, closed]
+    branches: [main, master]
 
 jobs:
   flowzone:
     name: Flowzone
     uses: product-os/flowzone/.github/workflows/flowzone.yml@master
+    # prevent duplicate workflow executions for pull_request and pull_request_target
+    if: |
+      (
+        github.event.pull_request.head.repo.full_name == github.repository &&
+        github.event_name == 'pull_request'
+      ) || (
+        github.event.pull_request.head.repo.full_name != github.repository &&
+        github.event_name == 'pull_request_target'
+      )
     secrets: inherit
     with:
-      rust_binaries: false
-      rust_toolchain: 1.67.1
+      rust_binaries: true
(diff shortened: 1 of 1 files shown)


.github/workflows/flowzone.yml
```diff
@@ -1,20 +1,29 @@
-name: Flowzone
```
klutchell, line 1: why did you remove this?

Cross.toml
```diff
@@ -1,23 +1,23 @@
...
+    "apt-get update && apt-get install --assume-yes libdbus-1-dev:$CROSS_DEB_ARCH",
 ]
 
 [target.i686-unknown-linux-gnu]
 pre-build = [
     "dpkg --add-architecture $CROSS_DEB_ARCH",
-    "apt-get update && apt-get install --assume-yes libdbus-1-dev:$CROSS_DEB_ARCH"
-]
+    "apt-get update && apt-get install --assume-yes libdbus-1-dev:$CROSS_DEB_ARCH",
+]
```
klutchell, line 23: Did anything change in this file besides formatting? I would prefer we not commit this file if functionally unchanged

.github/workflows/flowzone.yml
rahul-thakoor: i simply replicated the one we have at https://github.com/balena-os/takeover/blob/master/.github/workflows/flowzone.yml

Cross.toml
rahul-thakoor: only formatting. will leave it out

.github/workflows/flowzone.yml
klutchell: keep the name, it looks better in Action runs

rahul-thakoor: it's specified here https://github.com/balena-os/wifi-connect/blob/dda3275fe7b7cac68350c64d74d97c464af17c0a/.github/workflows/flowzone.yml#L12

rahul-thakoor: which is better? at the root or at the job level @klutchell ?

klutchell: Both will show up, look at the status checks below...

Do you prefer

`.github/workflows/flowzone.yml / Flowzone / Versioned source (pull_request)`

or `Flowzone / Flowzone / Versioned source (pull_request)`

The latter looks better to me, and it's what you get when you add a name to the top level of the workflow.

rahul-thakoor: I see, TIL! thanks

klutchell (approved):

PR #603, opened 2024-10-29
Do not publish the repo to crates.io

pipex:
The default is true. This prevents accidental publishing of the project

Change-type: patch


rahul-thakoor (approved):
Good catch

PR #646, opened 2026-01-23
fix(Dockerfile.template): add missing dependencies

aberenyi:
Namely `curl` and `ca-certificates`.


aberenyi:
This is esentially the same as #631, just checking whether the checks are working or not.

